PR #111, opened 2026-02-19
[Bugfix] adding completableFuture docs and minor bug fixes

### LINKED ISSUE #97: [Refactor]: issues with Phaser
### What would you like?

Currently we use phaser to synchronize the completion of an operation.

It has 3 phases
- RUNNING = 0
- COMPLETE = 1
- DONE

The lifecycle of a phaser is
- when an operation is created, a phaser is started (registered 1)
- when the operation is replayed and a completion state is received, the phaser will be completed (arriveAndAwait)
- when a completion state is received from checkpoint API, the phaser will be completed (arriveAndAwait)
- when the operation is waited (in `Operation.get()`),
   - if the phaser is completed (phase 1+), proceed without an action;
   - otherwise, the phaser will be registered with a new party (register), and waited (arriveAndAwait)

We have a few issues here
- #72 a race condition 
- #81 we allow the phasers to advance over COMPLETE. If we received duplicate completion updates from API, the phase could be 2, 3, ...
- The number of parties is unknown. We assume that there is only 1 party when the operation is replayed,  and 2 parties when an operation state is received from checkpoint API. But the assumption could be wrong. In complex situations, this could lead to dead locks (not enough arrive calls) or exceeding the phase range (too many arrive calls)

### Possible Implementation

A possible solution is to replace Phaser with CompletableFuture.

Assume we have a `CompletableFuture` field `completionFuture` per Operation
- when an operation is created, completionFuture is created (not completed)
- when the operation is replayed and a completion state is received, completionFuture will be completed.
- when a completion state is received from checkpoint API, completionFuture will be completed.
- when the operation is waited (in `Operation.get()`),
   - if the completionFuture is completed, proceed without an action;
   - otherwise, the completionFuture will be waited.

```
// in waitForOperationCompletion()
if (!completionFuture.isDone()) {
    // Operation not done yet
    var context = executionManager.getCurrentContext();
    // Deregister current context - allows suspension
    logger.debug(
            "get() on {} attempting to deregister context: {}",
            getType(),
            executionManager.getCurrentContext().contextId());
    deregisterActiveThread(context.contextId());

    // Block until operation completes
    logger.trace("Waiting for operation to finish {}", getOperationId());
    completionFuture.get();

    // Reactivate current thread
    registerActiveThread(context.contextId(), context.threadType());
}
```

With this approach, we will solve the current issues with Phaser, and have additional benefits:
- `allOf` and `anyOf` could be implemented easily and properly with the help of CompletableFuture. Currently `allOf` returns a list of values instead of a future.
- Updates can be directly passed from ExecutionManager to the operations via the future without operations calling ExecutionManager method

No

### Does this require an RFC?

No

### Additional Context

_No response_

zhongkechen:

### Issue Link, if available

Fix #97  

Follow-up actions on previous PR #103 

### Description

- Added more comments to CompletableFuture related code
- Added a new design doc ADR-003 for CompletableFuture based operation coordination
- Removed duplicate method `isTerminalStatus`
- Fixed a bug in ChildContextOperation that doesn't handle `SuspendExecutionException` properly, causing flaky test cases
- Fixed minor issues with runners
   - Added handling of errors in `CloudDurableTestRunner`
   - Fixed LocalDurableTestRunner gets stuck when skipTimer is true 
- Replaced some magic values in test cases with constants
- simplified thread local and active thread management
   - removed unset/reset thread local object from `waitForOperationCompletion`. Thread local object is set when the thread is started and it is recycled when the thread is terminated. 
   - removed thread type from active thread tracker. Only thread id is needed to track active threads.
   - renamed some methods to better reflect the purpose and avoid confusion. `Context` is renamed to `ThreadContext` to reflect that it's stored in thread local.

### Checklist

#### Unit Tests

#### Integration Tests

#### Examples

### DIFF
--- a/docs/adr/003-completable-future-based-coordination.md
+++ b/docs/adr/003-completable-future-based-coordination.md
@@ -1,4 +1,177 @@
 # ADR-003: CompletableFuture-Based Operation Coordination
 
-**Status:** Todo
+**Status:** Review
 **Date:** 2026-02-18
+
+## Context
+
+Currently, the SDK employs a Phaser-based mechanism for coordinating operations. The design is detailed in [ADR-002: Phaser-Based Operation Coordination](002-phaser-based-coordination.md).
+
+With this design, we can:
+ 
+- Register a thread when it begins and deregister it when it completes;
+- Block `DurableFuture::get()` calls until the operation completes;
+- Suspend execution when no registered thread exists.
+
+However, this design has a few issues:
+
+- We allow the Phasers to advance over predefined phase ranges (0 - RUNNING, 1 - COMPLETE). If we received duplicate completion updates from local runner or backend API, the phase could be advanced to 2, 3, and so on.
+- We assume that there is only one party during operation replay, and two parties when receiving an operation state from checkpoint API. We call Phaser `arriveAndAwaitAdvance` once or twice based on this assumption, but it could be incorrect. In complex scenarios, this could lead to a deadlock (not enough arrive calls) or exceeding the phase range (too many arrive calls).
+- The Phaser has higher complexity and cognitive overhead compared to other synchronization mechanisms.
+
+## Decision
+
+We will implement operation coordination using `CompletableFuture`.,
+
+### Threads
+
+Each piece of user code (e.g. the main Lambda function body, a step body, a child context body) runs in its own user thread from the user thread pool. 
+Execution manager tracks active running user threads. 
+When a new step or a new child context is created, a new thread is created and registered in execution manager. 
+When the step or the child context completes, the corresponding thread is deregistered from execution manager.
+When the user code is blocked on `DurableFuture::get()` or another synchronous durable operation (e.g., `wait()`), the caller thread is deregistered from execution manager. 
+When there is no registered thread in execution manager, the durable execution is suspended.
+
+A special SDK thread is created and managed by the SDK to make checkpoint API requests.
+
+### CompletableFuture
+
+The `CompletableFuture` is used to manage the completion of operations. It allows us to track the progress of operations and handle their completion in a more flexible and readable manner.
+
+Each durable operation has a `CompletableFuture` field.
+This field is used by user threads and the SDK thread communicate the completion of operations. 
+
+For example, when a context executes a step, the communication occurs as follows
+
+```mermaid
+sequenceDiagram
+    participant Context as Context Thread
+    participant Future as CompletableFuture
+    participant EM as Execution Manager
+    participant SDK as SDK Thread
+    participant Step as Step Thread
+
+    Note over Context: calling context.stepAsync()
+    Context->>Context: create StepOperation 
+    Context->>Future: create CompletableFuture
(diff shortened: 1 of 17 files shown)


phipag (approved):

zhongkechen:
The following is AI (GPT 5.5)'s opinion on ADR-002 Phaser vs ADR-003 CompletableFuture for reference:


I’d still choose **ADR-003 / `CompletableFuture`** for this SDK.

The short version: Durable operation coordination is fundamentally a **completion problem**, not a **barrier synchronization problem**. A step, wait, invoke, or callback eventually reaches a terminal state, and any blocked caller should resume after that terminal state is durably recorded. `CompletableFuture` models that directly. `Phaser` can be made to work, but it asks the implementation to encode lifecycle meaning into phase numbers and party counts, which is exactly the kind of cleverness that becomes fragile under replay, duplicate updates, and edge-case timing.

Why `CompletableFuture` is better here:

1. **The domain is one-shot completion.**  
   A durable operation needs to say: “I am complete; unblock anyone waiting on me.” `CompletableFuture.complete(...)` naturally supports that and is idempotent in spirit: duplicate completion attempts do not advance some hidden global phase. That matters because durable systems often see repeated state from replay, polling, retries, or backend responses.

2. **`Phaser` depends on assumptions that the system cannot always guarantee.**  
   ADR-003 calls this out well: the Phaser model assumes a certain number of parties and specific phase transitions. In a concurrent replay/checkpoint system, those assumptions are hard to maintain. If you under-arrive, you deadlock. If you over-arrive, you move past the intended phase. Both are nasty failure modes.

3. **`CompletableFuture` makes the coordination object match the operation.**  
   Each operation owns a future. Waiters attach to that operation. Completion happens when the operation is durably complete. That gives a cleaner mental model than “operation phase 0 means running, phase 1 means completing, phase 2 means safe to deregister,” especially for future contributors.

4. **The failure modes are easier to test and explain.**  
   The `CompletableFuture` design still has subtle ordering requirements, especially around re-registering a blocked user thread before the step thread deregisters. But that logic is localized. The current code’s synchronized block around future completion/waiter registration is easier to reason about than a multi-phase protocol spread across operation execution and checkpoint callbacks.

5. **It leaves the two-thread-pool idea intact.**  
   The valuable part of Philipp’s design was isolating user code from SDK/system work. `CompletableFuture` does not reject that; it gives the same architecture a simpler coordination primitive.

The best argument for ADR-002 is that `Phaser` explicitly models a multi-party handoff, and the two-phase protocol was trying to solve a real race: “don’t let the last worker deregister before the blocked caller is active again.” That is a serious problem, and the ADR-002 design shows thoughtful understanding of it. But I think `Phaser` is too powerful and too implicit for this case. It makes correctness depend on everyone participating in exactly the right phase choreography.

So my judgment: **ADR-003 is the better long-term design**, not because ADR-002 was careless, but because ADR-002 solved a hard problem with a synchronization primitive whose semantics did not quite fit durable replay. For SDK code that customers rely on, boring and directly modeled usually wins.

PR #122, opened 2026-02-25
Update pom file to support maven publish

wangyb-A:

### Description

- Parent:
  - Add gpg sign plugin
  - Add maven publish plugin
  - Add OSS information (url, licenses, developers, scm)
  - Add flattern plugin so that we can manage the information in the parent pom
- Children:
  - Include source file
  - Include javadoc file

### Checklist

### Testing

- Release to maven

#### Unit Tests

N/A

#### Integration Tests

N/A

#### Examples

N/A

### DIFF
--- a/sdk-integration-tests/pom.xml
+++ b/sdk-integration-tests/pom.xml
@@ -7,7 +7,7 @@
     <parent>
         <groupId>software.amazon.lambda.durable</groupId>
         <artifactId>aws-durable-execution-sdk-java-parent</artifactId>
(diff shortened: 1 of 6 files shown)


sdk-integration-tests/pom.xml
phipag: Similar comment like for examples. Should we skip this from publishing? These are the SDK internal integration tests. They are a standalone module to avoid a circulare dependency between the sdk and the sdk-testing modules since the integration tests use sdk-testing.

PR #124, opened 2026-02-26
fix/mavenurl

wangyb-A:

### Description

**Fix maven url in pom**
- We only set urls in parent POM before, flatten-pom-plugin will automatically append the artifact name in the URL, causing the url on maven to break.
- I added <url> and <scm> field in child pom so that it will not be overriden by the plugin.
- This fix will take effect when next release is published to maven.
- For the url used in <scm> section, please refer to: https://maven.apache.org/pom.html#scm

### Demo/Screenshots

N/A

### Checklist

### Testing

`mvn clean install`

Manually verified that in the `.flattened-pom.xml`, we are using

```
    <url>https://github.com/aws/aws-durable-execution-sdk-java</url>

    <scm>
        <connection>scm:git:https://github.com/aws/aws-durable-execution-sdk-java.git</connection>
        <developerConnection>scm:git:https://github.com/aws/aws-durable-execution-sdk-java.git</developerConnection>
        <url>https://github.com/aws/aws-durable-execution-sdk-java</url>
    </scm>
```

#### Unit Tests

N/A

#### Integration Tests

N/A

#### Examples

N/A

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
 
     <name>AWS Lambda Durable Execution SDK Parent</name>
     <description>Parent POM for AWS Lambda Durable Execution SDK and Examples</description>
-    <url>https://github.com/aws/aws-lambda-java-durable-execution-sdk</url>
+    <url>https://github.com/aws/aws-durable-execution-sdk-java</url>
 
     <licenses>
         <license>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -33,7 +33,7 @@
     </developers>
 
     <scm>
-        <url>https://github.com/aws/aws-lambda-java-durable-execution-sdk.git</url>
```
zhongkechen: This should be the git url

wangyb-A:
> have you tested this change? how does the link look like on sonatype?

Yes, I tested it locally. The link will be the same as that in the flattened-pom.xml. 
I don't think there's a way to test that with sonatype until we release a new version.

PR #201, opened 2026-03-12
feature: ConcurrencyOperation and Parallel

wangyb-A:

### Description

- Add parallel design from ParidelPooya @
- Add basic parallel operation

### Demo/Screenshots
iShot_2026-03-16_16 37 38

Note that the hierarchy is not correct, will be fixed in the follow up PRs
### Checklist

#### Unit Tests

#### Integration Tests

Have integration tests been written for these changes? Not yet

#### Examples

### DIFF
--- a/sdk/src/main/java/software/amazon/lambda/durable/operation/ConcurrencyOperation.java
+++ b/sdk/src/main/java/software/amazon/lambda/durable/operation/ConcurrencyOperation.java
@@ -57,7 +57,6 @@ public abstract class ConcurrencyOperation<T> extends BaseDurableOperation<T> {
     private ConcurrencyCompletionStatus completionStatus;
     private OperationIdGenerator operationIdGenerator;
 
-
     protected ConcurrencyOperation(
             OperationIdentifier operationIdentifier,
             TypeToken<T> resultTypeToken,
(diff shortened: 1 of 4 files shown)


docs/core/parallel.md
```diff
@@ -0,0 +1,143 @@
...
+    // AutoCloseable ensures join() is called
+    public void close();
+}
+```
+
+#### 3. DurableContext Integration
+Add single method to existing `DurableContext`:
+
+```java
+public ParallelContext parallel(ParallelConfig config);
```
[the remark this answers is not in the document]
wangyb-A: yeah, my point is if we return an instance of `ParallelOperation `.  We are going to keep both of the user interface and the concurrency operation logic in that instance.

Which is different from the original design, using a `ParallelContext` to represent the user interface, while using the ParallelOperation class for execution management

sdk/pom.xml
```diff
@@ -63,6 +63,12 @@
             <artifactId>mockito-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
+            <version>${slf4j.version}</version>
+            <scope>test</scope>
```
zhongkechen: why is this needed for test?

wangyb-A: It's for local debugging

Issue #39, opened 2026-01-30
[Feature]: design and implement `map`


zhongkechen:
### What would you like?

I would like to call `ctx.map` to run an operation over a list of items.

### Is this a breaking change?

No

### Does this require an RFC?

No

PR #229, opened 2026-03-18
feat(map): Fix map replay, wait-inside-map, and concurrency race conditions

ayushiahjolia:

### Issue Link, if available
https://github.com/aws/aws-durable-execution-sdk-java/issues/39

### Description
- Updated Map implementation to rely on ConcurrencyOperation. 
- Added tests for all map usages.

Follow-ups/Remaining Work
- If function returns null output, are we allowing checkpointing that?
- Is it okay to enqueue all items? or we want to start executing item as soon as possible.
- Doc updates in next CR, will remove dead code from Test as well.
- Thread issues during replay in LocalDurableTestRunner. 
- If minSuccessful is reached then remaining iterations should return `NOT_STARTED` or `SUCCESSFUL`?

### Demo/Screenshots
Screenshot 2026-03-19 at 4 36 00 PM

### Checklist

#### Unit Tests

Have unit tests been written for these changes? Updated.

#### Integration Tests

#### Examples

### DIFF
--- a/sdk/src/main/java/software/amazon/lambda/durable/operation/ConcurrencyOperation.java
+++ b/sdk/src/main/java/software/amazon/lambda/durable/operation/ConcurrencyOperation.java
@@ -282,7 +282,7 @@ public abstract class ConcurrencyOperation<T> extends BaseDurableOperation<T> {
     }
 
     private void handleComplete() {
-        synchronized (completionFuture) {
+        synchronized (this) {
             if (isOperationCompleted()) {
                 return;
             }
@@ -291,7 +291,6 @@ public abstract class ConcurrencyOperation<T> extends BaseDurableOperation<T> {
             } else {
                 handleFailure(completionStatus);
             }
-            completionFuture.complete(null);
         }
     }
 


ayushiahjolia:
When all map branches call wait(), every branch thread deregisters and suspendExecution() fires. But the parent thread sitting on completionFuture.join() in BaseConcurrentOperation.get() is stuck forever - no branch will ever complete to trigger onChildContextComplete and finalize the map's completionFuture. runUntilCompleteOrSuspend races completionFuture against executionExceptionFuture, so when suspension fires, the parent thread gets freed via SuspendExecutionException instead of blocking indefinitely.

sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java
```diff
@@ -209,8 +209,12 @@ protected Operation waitForOperationCompletion() {
...
         }
 
-        // Block until operation completes. No-op if the future is already completed.
-        completionFuture.join();
+        // Block until operation completes or execution suspends.
+        // Using runUntilCompleteOrSuspend races completionFuture against executionExceptionFuture,
+        // so when all active threads suspend (e.g., wait inside map branches), the
+        // SuspendExecutionException propagates and this thread is freed — preventing thread leaks
+        // on shared executor pools across invocations.
+        executionManager.runUntilCompleteOrSuspend(completionFuture).join();
```
zhongkechen: This is not the correct behavior. This will unblock all the threads waiting for completion of this operation when suspension.

sdk/src/main/java/software/amazon/lambda/durable/operation/BaseConcurrentOperation.java
```diff
@@ -85,9 +91,12 @@ protected BaseConcurrentOperation(
     protected void start() {
         sendOperationUpdateAsync(
                 OperationUpdate.builder().action(OperationAction.START).subType(subType.getValue()));
+        var parentThreadContext = getCurrentThreadContext();
         this.rootContext = getContext().createChildContext(getOperationId(), getName());
```
zhongkechen: Why do we initialize here?

sdk/src/main/java/software/amazon/lambda/durable/operation/BaseConcurrentOperation.java
```diff
@@ -299,32 +410,44 @@ public R get() {
...
     protected List<ChildContextOperation<?>> getBranches() {
-        return Collections.unmodifiableList(branches);
+        synchronized (lock) {
+            return Collections.unmodifiableList(new ArrayList<>(branches));
+        }
     }
 
     protected CompletionReason getCompletionReason() {
-        return completionReason;
+        synchronized (lock) {
```
zhongkechen: Why do we need a lock to access a field? Same below

sdk/src/main/java/software/amazon/lambda/durable/operation/WaitOperation.java
```diff
@@ -71,6 +71,10 @@ private void pollForWaitExpiration() {
                 && existing.waitDetails().scheduledEndTimestamp() != null) {
             remainingWaitTime =
                     Duration.between(Instant.now(), existing.waitDetails().scheduledEndTimestamp());
+            // If the wait has already elapsed, poll immediately with a minimal positive interval
+            if (remainingWaitTime.isNegative() || remainingWaitTime.isZero()) {
+                remainingWaitTime = Duration.ofMillis(1);
+            }
```
zhongkechen: Is this necessary? We don't have this check in other places calling `pollForOperationUpdates`

sdk/src/main/java/software/amazon/lambda/durable/operation/MapOperation.java
```diff
@@ -5,143 +5,285 @@
...
+        // All items finished (no pending, no running) — complete with ALL_COMPLETED
+        if (isAllItemsFinished()) {
+            completionStatus = ConcurrencyCompletionStatus.ALL_COMPLETED;
+            return true;
         }
 
-        return mapResult;
+        return false;
+    }
+
```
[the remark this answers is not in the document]
ayushiahjolia: I think for parallel allCompleted means no failures right? But for Map it includes failed iterations as well.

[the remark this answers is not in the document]
wangyb-A: Ah that might be a miss of parallel impl, all complete should also include failure (minSuccess reached)
I'll fix this later

PR #242, opened 2026-03-20
refactor(map): Fix null results, abstract completion logic

ayushiahjolia:

### Issue Link, if available
https://github.com/aws/aws-durable-execution-sdk-java/issues/39

### Description
- Fix NPE in ChildContextOperation.checkpointSuccess when map iterations return null by adding a null guard on the serialized payload before calling getBytes().
- Make canComplete() and validateItemCount() abstract in ConcurrencyOperation so Map and Parallel each own their completion semantics. Move Parallel-specific logic (fail-fast on any failure when minSuccessful == -1) into ParallelOperation. - Remove unused minSuccessful/toleratedFailureCount/failureRateThreshold fields from the base class.
- Remove dead runUntilCompleteOrSuspend stub from BaseDurableOperationTest.
- Add testMapWithNullResults integration test. 
- Update design.md with ConcurrencyOperation, MapOperation, ParallelOperation, and ChildContextOperation in the architecture diagram and package structure.

### Demo/Screenshots
N/A

### Checklist

#### Unit Tests

Have unit tests been written for these changes? Updated required tests.

#### Integration Tests

Have integration tests been written for these changes? Added missing tests.

#### Examples


zhongkechen (dismissed):

sdk/src/main/java/software/amazon/lambda/durable/operation/ParallelOperation.java
```diff
@@ -118,4 +115,38 @@ public T get() {
...
+    }
+
+    @Override
+    protected boolean canComplete() {
+        int succeeded = getSucceededCount();
+        int failed = getFailedCount();
+
+        // If we've met the minimum successful count, we're done
+        if (minSuccessful != -1 && succeeded >= minSuccessful) {
+            completionStatus = ConcurrencyCompletionStatus.MIN_SUCCESSFUL_REACHED;
```
zhongkechen: `canComplete` as the name implies should be read only. `completionStatus` can be determined later or returned as a return value. Mutating a field here would be surprising and require a concurrency control.

sdk/src/main/java/software/amazon/lambda/durable/operation/ConcurrencyOperation.java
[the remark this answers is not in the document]
ayushiahjolia: Yeah, replace it.

sdk/src/main/java/software/amazon/lambda/durable/operation/ParallelOperation.java
ayushiahjolia: Updating it.

Issue #218, opened 2026-03-16
[Feature]: convenient methods for UUID/timestamp/random


zhongkechen:
### What would you like?

Create convenient methods in DurableContext to create steps wrapping UUID, timestamp and etc.
- ctx.uuid() - UUID
- ctx.now() - current Instant
- ctx.random() - generate a random number between 0 and 1

### Is this a breaking change?

No

### Does this require an RFC?

No

PR #343, opened 2026-04-22
feat: withRetry helper

### LINKED ISSUE #355: withRetry
implement the new withRetry helper

https://github.com/aws/aws-durable-execution-sdk-js/pull/505

SilanHe:

### Issue Link, if available

Fixes: #355 

### Description

Add a public `WithRetry` helper that retries any durable operation (`waitForCallback`, `invoke`, `waitForCondition`, etc.) end-to-end with configurable backoff. Today users have to hand-roll a while loop with manual `context.wait()` backoff calls. This helper provides a single, replay-safe, well-tested primitive for that pattern, reusing the exact same `RetryStrategy` shape developers already know from `StepConfig`.

**New files:**

- `withRetryConfig` (`sdk/.../config/`) — Builder-pattern config holding the `RetryStrategy` (same type as `StepConfig`) and `wrapInChildContext` (defaults to `true`). If RetryStrategy is not specified a default RetryStrategy is used.

**Key behaviors:**

- Reuses the existing `RetryStrategy` / `RetryDecision` types — zero new retry concepts.
- Replay-safe by construction: every side-effect in the loop is a durable operation (`operation.execute()` calls durable primitives, `context.wait()` handles backoff).
- Correctly propagates `SuspendExecutionException` and `UnrecoverableDurableExecutionException` without retrying — these are internal SDK control flow signals.
- withRetry and withRetryAsync(context, name, operation, config)` — wraps in `runInChildContext` by default for clean execution history grouping.

### Demo/Screenshots

N/A

### Checklist

#### Unit Tests

yes

#### Integration Tests

yes

#### Examples

Yes.

yes


examples/src/main/java/software/amazon/lambda/durable/examples/callback/RetryWaitForCallbackExample.java
```diff
@@ -0,0 +1,50 @@
...
+    public String handleRequest(ApprovalRequest input, DurableContext context) {
+        // Step 1: Prepare the approval request
+        var prepared = context.step(
+                "prepare",
+                String.class,
+                stepCtx -> "Approval for: " + input.description() + " ($" + input.amount() + ")");
+
+        // Step 2: waitForCallback with retry — if the external system fails, try again with a fresh callback
+        var approvalResult = WithRetryHelper.withRetry(
+                context,
```
zhongkechen: If `withRetry` requires a durable context parameter, why don't we put this method into `DurableContext`?

sdk/src/main/java/software/amazon/lambda/durable/util/WithRetryHelper.java
```diff
@@ -0,0 +1,140 @@
...
+     *
+     * @param <T> the result type
+     * @param context the durable context
+     * @param name operation name (used for child context and backoff wait names)
+     * @param operation the retryable operation — receives the context and 1-based attempt number
+     * @param config retry configuration including the retry strategy
+     * @return the operation result
+     */
+    @SuppressWarnings("unchecked")
+    public static <T> T withRetry(DurableContext context, String name, WithRetry<T> operation, WithRetryConfig config) {
```
zhongkechen: Where is async version `withRetryAsync` of it? `withRetryAsync` will return a DurableFuture and users can combine that with `anyOf`/`allOf` or just kick off an operation with retry in the background.

examples/src/main/java/software/amazon/lambda/durable/examples/callback/RetryWaitForCallbackExample.java
```diff
@@ -0,0 +1,50 @@
...
+                (ctx, attempt) -> ctx.waitForCallback(
+                        "approval-" + attempt, String.class, (callbackId, stepCtx) -> stepCtx.getLogger()
+                                .info("Attempt {}: sending callback {} to approval system", attempt, callbackId)),
```
zhongkechen, line 40: Does this supplier have to contain only one operation? How do we ensure and validate this?

sdk/src/main/java/software/amazon/lambda/durable/config/WithRetryConfig.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /**
+     * Whether to wrap the retry loop in {@code runInChildContext} so all attempts are grouped under a single named
+     * operation in execution history. Only applies when a name is provided to the named form of {@code withRetry}.
+     * Defaults to {@code true}.
+     *
+     * @return true if child-context wrapping is enabled
+     */
+    public boolean wrapInChildContext() {
```
zhongkechen, line 44: In which scenarios `wrapInChildContext` is mandate?

sdk/src/main/java/software/amazon/lambda/durable/config/WithRetryConfig.java
```diff
@@ -0,0 +1,103 @@
...
+
+        /**
+         * Builds the {@link WithRetryConfig} instance.
+         *
+         * @return a new config with the configured options
+         * @throws IllegalArgumentException if retryStrategy is not set
+         */
+        public WithRetryConfig build() {
+            if (retryStrategy == null) {
+                throw new IllegalArgumentException("retryStrategy is required");
```
zhongkechen: All existing configurations are optional. I think we should keep the convention for this new `WithRetryConfig`

zhongkechen (changes requested):

sdk/src/main/java/software/amazon/lambda/durable/util/WithRetryHelper.java
```diff
@@ -0,0 +1,140 @@
...
+        Objects.requireNonNull(context, "context cannot be null");
+        Objects.requireNonNull(name, "name cannot be null");
+        Objects.requireNonNull(operation, "operation cannot be null");
+        Objects.requireNonNull(config, "config cannot be null");
+
+        if (config.wrapInChildContext()) {
+            return (T) context.runInChildContext(
+                    name, new TypeToken<Object>() {}, childCtx -> executeRetryLoop(childCtx, name, operation, config));
```
zhongkechen: We could have a new subtype for this child context.

sdk/src/main/java/software/amazon/lambda/durable/config/WithRetryConfig.java
SilanHe: this is a user setting, so it's up to the user whether they want a child context or not

sdk/src/main/java/software/amazon/lambda/durable/util/WithRetryHelper.java
SilanHe: ah good catch, I missed this

sdk/src/main/java/software/amazon/lambda/durable/util/WithRetryHelper.java
SilanHe: what do you mean?

zhongkechen: What's the subtype of this CONTEXT operation?

examples/src/main/java/software/amazon/lambda/durable/examples/callback/RetryWaitForCallbackExample.java
SilanHe: no I don't think it has to contain only one operation

zhongkechen: In the case of multiple operations, does wrapInChildContext have to be true?

examples/src/main/java/software/amazon/lambda/durable/examples/callback/RetryWaitForCallbackExample.java
SilanHe: To be honest, I don't have a strong opinion on this. The current approach is the approach that the Typescript SDK took, to have it included as "util" rather than editing the `DurableContext`. This isn't a "primitive" operation, but more something that composes existing operations together. In my opinion, it's a little more clear to define it separately.

sdk/src/main/java/software/amazon/lambda/durable/config/WithRetryConfig.java
SilanHe: Are we sure we want to do that in this specific case? How we would want to proceed in the case of a withRetry with null retryStrategy? Not retry? or do we want to include a default retryStrategy?

examples/src/main/java/software/amazon/lambda/durable/examples/callback/RetryWaitForCallbackExample.java
zhongkechen: Utils like allOf are also in DurableContext. Lets keep this in that as well

sdk/src/main/java/software/amazon/lambda/durable/config/WithRetryConfig.java
zhongkechen: We can have a default retry strategy, just like the step or waitForCondition operations

examples/src/main/java/software/amazon/lambda/durable/examples/callback/RetryWaitForCallbackExample.java
SilanHe: ok

examples/src/main/java/software/amazon/lambda/durable/examples/callback/RetryWaitForCallbackExample.java
SilanHe: no, I don't think so.

examples/src/main/java/software/amazon/lambda/durable/examples/callback/RetryWaitForCallbackExample.java
SilanHe: I did a quick search and `allOf` is in `DurableFuture` not in `DurableContext`. Do you still want to move this `withRetry` to `DurableContext`?

sdk/src/main/java/software/amazon/lambda/durable/config/WithRetryConfig.java
SilanHe: ok, that sounds good

examples/src/main/java/software/amazon/lambda/durable/examples/callback/RetryWaitForCallbackExample.java
zhongkechen: Yeah, I still think it's better to keep the user interfacing classes fewer the better. DurableContext is always the main entrance where users look for durable operations, no matter it's a real operation or a virtual one, or a utility like #218.

sdk/src/main/java/software/amazon/lambda/durable/util/WithRetryHelper.java
```diff
@@ -0,0 +1,199 @@
...
+     * @param config retry configuration including the retry strategy
+     * @return a future representing the operation result
+     */
+    public static <T> DurableFuture<T> withRetryAsync(
+            DurableContext context, WithRetry<T> operation, WithRetryConfig config) {
+        Objects.requireNonNull(context, "context cannot be null");
+        Objects.requireNonNull(operation, "operation cannot be null");
+        Objects.requireNonNull(config, "config cannot be null");
+
+        return new CompletedDurableFuture<>(executeRetryLoop(context, null, operation, config));
```
zhongkechen: I'm afraid this CompletedDurableFuture doesn't work well, especially with DurableFuture::anyOf

sdk/src/main/java/software/amazon/lambda/durable/config/WithRetryConfig.java
zhongkechen: The operations must be generated in deterministic order. I think combining wrapInChildContext=false and withRetryAsync can lead to indeterministic order of operations.

sdk/src/main/java/software/amazon/lambda/durable/util/WithRetryHelper.java
SilanHe: you mean if we have an withRetryAsync?

sdk/src/main/java/software/amazon/lambda/durable/config/WithRetryConfig.java
SilanHe: ah yeah that's true, I don't know if we took this into account on the Typescript SDK.

sdk/src/main/java/software/amazon/lambda/durable/util/WithRetryHelper.java
zhongkechen: CONTEXT operation of Map has MAP subtype, iterations of map have MAP_ITERATION subtype, waitForCallback has WAIT_FOR_CALLBACK subtype, etc. So I think we can create a new subtype for this use case.

sdk/src/main/java/software/amazon/lambda/durable/config/WithRetryConfig.java
zhongkechen: I think when wrapInChildContext is false, a **virtual** context should be used to make sure the operations from this virtual context don't interfere with the other operations from the parent context.

sdk/src/main/java/software/amazon/lambda/durable/util/WithRetryHelper.java
zhongkechen: A new subtype `WITH_RETRY` can be used here when checkpointing the context operation, so that users can see the context's purpose properly on console.

sdk/src/main/java/software/amazon/lambda/durable/util/WithRetryHelper.java
zhongkechen: Created an issue for this: https://github.com/aws/aws-durable-execution-sdk-java/issues/357

examples/src/main/java/software/amazon/lambda/durable/examples/callback/RetryWaitForCallbackExample.java
SilanHe: ok got it

sdk/src/main/java/software/amazon/lambda/durable/util/WithRetryHelper.java
SilanHe: oh I understand now, `OperationSubType` is the enum you're referring to. I can do that.

sdk/src/main/java/software/amazon/lambda/durable/util/WithRetryHelper.java
SilanHe: Thank you

sdk/src/main/java/software/amazon/lambda/durable/config/WithRetryConfig.java
SilanHe: and this is for the withRetry() case, we still need the actual child context for the withRetryAsync() ?

SilanHe: I've added the use of virtual child context in all cases where wrapInChildContext is false

sdk/src/main/java/software/amazon/lambda/durable/DurableContext.java
```diff
@@ -727,6 +729,76 @@ <T> DurableFuture<T> waitForConditionAsync(
...
+     * {@link WithRetryConfig#wrapInChildContext()} is enabled, the child context is checkpointed (persisted) so all
+     * attempts are grouped under a single named operation in execution history. Otherwise, a virtual child context is
+     * used — no checkpointing overhead, but the child re-executes on replay.
+     *
+     * @param <T> the result type
+     * @param operation the retryable operation — receives the context and 1-based attempt number
+     * @param config retry configuration including the retry strategy and child context wrapping
+     * @return the operation result
+     */
+    <T> T withRetry(WithRetry<T> operation, WithRetryConfig config);
```
zhongkechen: `name` always exists in the other durable methods.

sdk/src/main/java/software/amazon/lambda/durable/context/DurableContextImpl.java
```diff
@@ -366,6 +392,137 @@ public <T> DurableFuture<T> waitForConditionAsync(
...
+    // =============== withRetry ================
+
+    private static final Duration DEFAULT_BACKOFF_DELAY = Duration.ofSeconds(1);
+    private static final String BACKOFF_SUFFIX = "-backoff-";
+    private static final String ANONYMOUS_CHILD_CONTEXT_NAME = "retry";
+    private static final String ANONYMOUS_BACKOFF_PREFIX = "retry-backoff-";
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public <T> T withRetry(String name, WithRetry<T> operation, WithRetryConfig config) {
```
zhongkechen: Isn't `withRetry` just `withRetryAsync` + `.get`

sdk/src/main/java/software/amazon/lambda/durable/context/DurableContextImpl.java
```diff
@@ -248,7 +272,9 @@ private <T> DurableFuture<T> runInChildContextAsync(
                 func,
                 resultType,
                 config,
-                this);
+                this,
+                isVirtual,
+                null);
```
zhongkechen: I'm afraid this doesn't work when the parent is virtual like in a virtual branch. Can you add a case to test this?

sdk/src/main/java/software/amazon/lambda/durable/context/DurableContextImpl.java
SilanHe: I thought that's what it was but then we talked about the need for a runInChildContext for the async for no matter what. Can the withRetryAsync function with a virtual context instead?

sdk/src/main/java/software/amazon/lambda/durable/DurableContext.java
SilanHe: do we want to enforce the name parameter in Java? It's optional in Typescript: https://github.com/aws/aws-durable-execution-sdk-js/blob/main/packages/aws-durable-execution-sdk-js/src/utils/with-retry/index.ts#L205

sdk/src/main/java/software/amazon/lambda/durable/context/DurableContextImpl.java
zhongkechen: `runInChildContext`  is needed for both async and non-async. All the combinations should work: `wrapInChildContext` (true or false) with `withRetry` or `withRetryAsync`

SilanHe: so 

withRetry wrapInChildContext=true -> runInChildContext with isVirtual=false
withRetry wrapInChildContext=false -> runInChildContext with isVirtual=true
withRetryAsync wrapInChildContext=true -> runInChildContext with isVirtual=false
withRetryAsync wrapInChildContext=false -> runInChildContext with isVirtual=true

?

zhongkechen: Yeah, always use the private `runInChildContextAsync` method

sdk/src/main/java/software/amazon/lambda/durable/DurableContext.java
zhongkechen: It's also optional in Java but users have to pass `null`. We wanted to avoid too many overloaded methods and also encourage users to always specify name for operations

sdk/src/main/java/software/amazon/lambda/durable/context/DurableContextImpl.java
zhongkechen: Fixed in: https://github.com/aws/aws-durable-execution-sdk-java/pull/363

sdk/src/main/java/software/amazon/lambda/durable/DurableContext.java
SilanHe: oh I see, ok I'll remove the extra function overload

sdk/src/main/java/software/amazon/lambda/durable/context/DurableContextImpl.java
SilanHe: thanks!

sdk/src/main/java/software/amazon/lambda/durable/DurableContext.java
```diff
@@ -727,6 +729,50 @@ <T> DurableFuture<T> waitForConditionAsync(
...
+     * used — no checkpointing overhead, but the child re-executes on replay.
+     *
+     * @param <T> the result type
+     * @param name operation name (used for backoff wait names, and as the child context name when wrapping); pass
+     *     {@code null} for an anonymous retry whose backoff waits use default names
+     * @param operation the retryable operation — receives the context and 1-based attempt number
+     * @param config retry configuration including the retry strategy and child context wrapping
+     * @return the operation result
+     */
+    default <T> T withRetry(String name, WithRetry<T> operation, WithRetryConfig config) {
```
zhongkechen: I thought you've noticed my previous comment that config can be optional. So we need an overloaded version without config.

SilanHe: ah no I did not. I can do that

sdk/src/main/java/software/amazon/lambda/durable/model/WithRetry.java
```diff
@@ -0,0 +1,27 @@
...
+public interface WithRetry<T> {
+
+    /**
+     * Executes the durable operation.
+     *
+     * @param context the durable context to use for durable operations
+     * @param attempt the current attempt number (1-based: first attempt is 1)
+     * @return the operation result
+     */
+    T execute(DurableContext context, int attempt);
```
zhongkechen: We didn't name any functional interfaces previously. We simply use `BiFunction` if the method takes two parameters. And context is always the last parameter in the previous scenarios.

SilanHe: ok, I'll change it to BiFunction instead.

examples/src/main/java/software/amazon/lambda/durable/examples/callback/RetryWaitForCallbackExample.java
SilanHe: We discussed offline, while wrapInChildContext can be false in the case of multiple operations, we will always use at least a **virtual** context to ensure that the potentially async code execution is durable across invocations.

zhongkechen (approved):

PR #349, opened 2026-04-24
feat(sdk): Treat KMS exceptions as non-retryable customer errors

denzyll:
### Issue Link, if available

N/A

### Description

KMS exceptions from Lambda (e.g., KMSAccessDeniedException, KMSDisabledException) arrive as 502 errors during
CheckpointDurableExecution and GetDurableExecutionState API calls. Unlike other 5xx errors, we do not want Durable Functions to retry invocations for these errors, as they would never self-resolve.

Add DurableApiErrorClassifier in the exception package with a NON_RETRYABLE_ERROR_CODES set and classifyException method. Use it in CheckpointManager to wrap both checkpoint() and getExecutionState() calls so that KMS errors return UnrecoverableDurableExecutionException with status FAILED immediately. The classifier also handles 4xx (non-429) as non-retryable and Invalid Checkpoint Token as retryable, matching the JS SDK's classifyCheckpointError logic.

The classifyException method could be further updated to return a new UnrecoverableDurableInvocationException for retryable Durable API errors (e.g., 429, 5xx), allowing the SDK to distinguish them from user code errors and crash the invocation for retry instead of misclassifying them as step failures.

### Checklist

- [X] I have filled out every section of the PR template
- [X] I have thoroughly tested this change

#### Unit Tests

Testing — parameterized unit tests for all four KMS exceptions with realistic error messages, plus tests for 4xx, Invalid Checkpoint Token, non-token InvalidParameterValueException, 429, 500, non-KMS 502, error detail preservation, and two wiring tests verifying CheckpointManager delegates to the classifier for both API paths.

#### Integration Tests

No.

#### Examples

N/A

### DIFF
--- a/sdk/src/main/java/software/amazon/lambda/durable/execution/CheckpointManager.java
+++ b/sdk/src/main/java/software/amazon/lambda/durable/execution/CheckpointManager.java
@@ -14,10 +14,12 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Consumer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.awscore.exception.AwsServiceException;
 import software.amazon.awssdk.services.lambda.model.CheckpointUpdatedExecutionState;
 import software.amazon.awssdk.services.lambda.model.Operation;
 import software.amazon.awssdk.services.lambda.model.OperationUpdate;
 import software.amazon.lambda.durable.DurableConfig;
+import software.amazon.lambda.durable.exception.DurableApiErrorClassifier;
 import software.amazon.lambda.durable.retry.PollingStrategies;
 import software.amazon.lambda.durable.retry.PollingStrategy;
 
@@ -163,14 +165,18 @@ class CheckpointManager {
         var nextMarker = checkpointUpdatedExecutionState.nextMarker();
         while (nextMarker != null && !nextMarker.isEmpty()) {
             var startTime = System.nanoTime();
-            var response = config.getDurableExecutionClient()
-                    .getExecutionState(durableExecutionArn, checkpointToken, nextMarker);
-            logger.debug(
-                    "Durable getExecutionState API called (latency={}ns): {}.",
-                    System.nanoTime() - startTime,
-                    response);
-            operations.addAll(response.operations());
-            nextMarker = response.nextMarker();
+            try {
+                var response = config.getDurableExecutionClient()
+                        .getExecutionState(durableExecutionArn, checkpointToken, nextMarker);
+                logger.debug(
+                        "Durable getExecutionState API called (latency={}ns): {}.",
+                        System.nanoTime() - startTime,
+                        response);
+                operations.addAll(response.operations());
+                nextMarker = response.nextMarker();
+            } catch (AwsServiceException e) {
+                throw DurableApiErrorClassifier.classifyException(e);
+            }
         }
         return operations;
     }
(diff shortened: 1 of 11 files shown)


sdk/src/main/java/software/amazon/lambda/durable/util/DurableApiErrorClassifier.java
```diff
@@ -0,0 +1,111 @@
...
+ * <p>Execution-level errors throw {@link UnrecoverableDurableExecutionException} to terminate the execution
+ * immediately. These represent permanent customer-side issues (e.g., KMS key misconfiguration) that will not
+ * self-resolve on retry.
+ *
+ * <p>Invocation-level errors are allowed to propagate, crashing the current Lambda invocation so the backend can retry
+ * with a fresh invocation.
+ *
+ * <p>To add a new non-retryable error, add its error code to {@link #NON_RETRYABLE_ERROR_CODES}.
+ */
+public final class DurableApiErrorClassifier {
```
zhongkechen, line 19: Can you put this class under `util` folder together with `ExceptionHelper` in the same place?

sdk/src/main/java/software/amazon/lambda/durable/execution/CheckpointManager.java
```diff
@@ -163,14 +165,18 @@ List<Operation> fetchAllPages(CheckpointUpdatedExecutionState checkpointUpdatedE
...
+                var response = config.getDurableExecutionClient()
+                        .getExecutionState(durableExecutionArn, checkpointToken, nextMarker);
+                logger.debug(
+                        "Durable getExecutionState API called (latency={}ns): {}.",
+                        System.nanoTime() - startTime,
+                        response);
+                operations.addAll(response.operations());
+                nextMarker = response.nextMarker();
+            } catch (AwsServiceException e) {
+                throw DurableApiErrorClassifier.classifyException(e);
```
zhongkechen, line 178: Don't see why this is necessary at this moment in this place. Anything thrown from here will eventually terminate the execution

zhongkechen: `classifyException` can throw a retryable exception with this change: https://github.com/aws/aws-durable-execution-sdk-java/issues/351

denzyll: > Don't see why this is necessary at this moment in this place. Anything thrown from here will eventually terminate the execution

It's so we could update `DurableApiErrorClassifier` in a follow-up PR to return a new `UnrecoverableDurableInvocationException` for retryable Durable API errors that should cause the SDK to retry the invocation instead of failing the execution immediately:

> The classifyException method could be further updated to return a new UnrecoverableDurableInvocationException for retryable Durable API errors (e.g., 429, 5xx), allowing the SDK to distinguish them from user code errors and crash the invocation for retry instead of misclassifying them as step failures.

> classifyException can throw a retryable exception with this change: #351

@zhongkechen implemented the follow-up PR here, which I use in `DurableApiErrorClassifier`. Instead of creating a new `UnrecoverableDurableInvocationException`, we continue to use `UnrecoverableDurableExecutionException` but with a `retryable` option.

sdk/src/main/java/software/amazon/lambda/durable/util/DurableApiErrorClassifier.java
denzyll: Done!

sdk/src/main/java/software/amazon/lambda/durable/execution/CheckpointManager.java
zhongkechen: typo: `UnrecoverableDurableExecutionException`

zhongkechen (approved):

PR #350, opened 2026-04-24
[feature]: add retryable exception

### LINKED ISSUE #351: [Feature]: retry the retryable exception
### What would you like?

When durable backend throws a retryable exception, and all the retry attempts in Lambda client fail, I would like the execution to throw a retryable exception so that the backend will retry the invocation

Currently in this situation, the execution will fail immediately

### Possible Implementation

_No response_

No

### Does this require an RFC?

No

### Additional Context

_No response_

zhongkechen:

### Issue Link, if available

Fixes: https://github.com/aws/aws-durable-execution-sdk-java/issues/351

### Description

- add retryable UnrecoverableDurableExecutionException
- throw the exception if retryable so that the backend can retry

### Checklist

#### Unit Tests

#### Integration Tests

#### Examples

### DIFF
--- a/sdk/src/main/java/software/amazon/lambda/durable/exception/UnrecoverableDurableExecutionException.java
+++ b/sdk/src/main/java/software/amazon/lambda/durable/exception/UnrecoverableDurableExecutionException.java
@@ -7,14 +7,25 @@ import software.amazon.awssdk.services.lambda.model.ErrorObject;
 /** Exception thrown when the execution is not recoverable. The durable execution will be immediately terminated. */
 public class UnrecoverableDurableExecutionException extends DurableExecutionException {
     private final ErrorObject errorObject;
+    private final boolean retryable;
 
-    public UnrecoverableDurableExecutionException(ErrorObject errorObject) {
+    public UnrecoverableDurableExecutionException(ErrorObject errorObject, boolean retryable) {
         super(errorObject.errorMessage());
         this.errorObject = errorObject;
+        this.retryable = retryable;
+    }
+
+    public UnrecoverableDurableExecutionException(ErrorObject errorObject) {
+        this(errorObject, false);
     }
 
     /** Returns the error details for this unrecoverable exception. */
     public ErrorObject getErrorObject() {
         return errorObject;
+    }
+
+    /** Returns true if the execution can be retried. */
+    public boolean isRetryable() {
+        return retryable;
     }
 }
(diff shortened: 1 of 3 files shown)


sdk/src/main/java/software/amazon/lambda/durable/exception/UnrecoverableDurableExecutionException.java
```diff
@@ -7,14 +7,25 @@
 /** Exception thrown when the execution is not recoverable. The durable execution will be immediately terminated. */
 public class UnrecoverableDurableExecutionException extends DurableExecutionException {
     private final ErrorObject errorObject;
+    private final boolean retryable;
 
-    public UnrecoverableDurableExecutionException(ErrorObject errorObject) {
+    public UnrecoverableDurableExecutionException(ErrorObject errorObject, boolean retryable) {
```
denzyll, line 12: The exception `UnrecoverableDurableExecutionException` is for unrecoverable executions, but the new retryable option contradicts this purpose. Instead, should we create a new `UnrecoverableDurableInvocationException` for retryable invocations, like how we have `CheckpointUnrecoverableInvocationError`/`UnrecoverableInvocationError` in the JS SDK?

sdk/src/main/java/software/amazon/lambda/durable/exception/UnrecoverableDurableExecutionException.java
```diff
@@ -7,14 +7,25 @@
 /** Exception thrown when the execution is not recoverable. The durable execution will be immediately terminated. */
```
denzyll, line 7: We should update this line to explain the retryable behaviour.

sdk/src/main/java/software/amazon/lambda/durable/exception/UnrecoverableDurableExecutionException.java
zhongkechen: Every exception in Java is named DurableExecutionException. It doesn't necessarily mean it will fail the execution.

sdk/src/main/java/software/amazon/lambda/durable/exception/UnrecoverableDurableExecutionException.java
zhongkechen: Yeah, the invocation will be immediately terminated. The SDK sometimes uses execution where it actually means invocation.

denzyll (approved):

PR #363, opened 2026-04-30
[feature]: add virtual to child context config

### LINKED ISSUE #360: [Feature]: expose virtual context to users
### What would you like?

I would like to create virtual context with `runInChildContext` method. Currently virtual context is only supported in concurrency operations when nestingType=FLAT

### Possible Implementation

_No response_

No

### Does this require an RFC?

No

### Additional Context

_No response_

zhongkechen:

### Issue Link, if available

- fixes: #360 

### Description

- added `isVirtual` to `RunInChildContextConfig` so that users can start a virtual context with no checkpoint overhead.
- recursively get the parent operation id until it's non-virtual. Currently it's always getting the direct parent for a virtual context because in map/parallel the direct parent is always non-virtual.

### Demo/Screenshots

image

### Checklist

#### Unit Tests

#### Integration Tests

#### Examples

### DIFF
--- a/sdk-integration-tests/src/test/java/software/amazon/lambda/durable/ChildContextIntegrationTest.java
+++ b/sdk-integration-tests/src/test/java/software/amazon/lambda/durable/ChildContextIntegrationTest.java
@@ -91,7 +91,7 @@ class ChildContextIntegrationTest {
         assertEquals(ExecutionStatus.FAILED, result.getStatus());
         assertEquals(1, childExecutionCount.get());
 
-        // Second run - replays, should throw same exception without re-executing
+        // Second run - replays, should throw same exception with re-executing
         result = runner.run("test");
         assertEquals(ExecutionStatus.FAILED, result.getStatus());
         assertTrue(result.getError().isPresent());
@@ -121,7 +121,7 @@ class ChildContextIntegrationTest {
         assertEquals(ExecutionStatus.FAILED, result.getStatus());
         assertEquals(1, childExecutionCount.get());
 
-        // Second run - replays, should throw same exception without re-executing
+        // Second run - replays, should throw same exception with re-executing
         result = runner.run("test");
         assertEquals(ExecutionStatus.FAILED, result.getStatus());
         assertTrue(result.getError().isPresent());
--- a/sdk-integration-tests/src/test/java/software/amazon/lambda/durable/ChildContextIntegrationTest.java
+++ b/sdk-integration-tests/src/test/java/software/amazon/lambda/durable/ChildContextIntegrationTest.java
@@ -8,6 +8,7 @@ import java.time.Duration;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.junit.jupiter.api.Test;
 import software.amazon.awssdk.services.lambda.model.OperationType;
+import software.amazon.lambda.durable.config.RunInChildContextConfig;
 import software.amazon.lambda.durable.model.ExecutionStatus;
 import software.amazon.lambda.durable.testing.LocalDurableTestRunner;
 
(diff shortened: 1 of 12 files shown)


sdk/src/main/java/software/amazon/lambda/durable/config/RunInChildContextConfig.java
```diff
@@ -60,6 +67,17 @@ public Builder serDes(SerDes serDes) {
             return this;
         }
 
+        /**
+         * Sets whether the context operation will be checkpointed (default: false).
```
SilanHe: nit: this is a little misleading. The default is that isVirtual=false which means that the context operation will be checkpointed right?

sdk/src/main/java/software/amazon/lambda/durable/context/DurableContextImpl.java
```diff
@@ -407,6 +407,6 @@ private String nextOperationId() {
      * @return the parent of this context if virtual, otherwise this context id
      */
     public String getParentId() {
-        return isVirtual ? parentContext.getContextId() : getContextId();
```
SilanHe, line 410: was this a bug?

zhongkechen: no, the direct parent is always non-virtual in existing use cases (parallel/map)

sdk/src/main/java/software/amazon/lambda/durable/config/RunInChildContextConfig.java
zhongkechen: Good catch. Updated

sdk-integration-tests/src/test/java/software/amazon/lambda/durable/ChildContextIntegrationTest.java
```diff
@@ -42,6 +43,34 @@ void childContextResultSurvivesReplay() {
...
+                        },
+                        RunInChildContextConfig.builder().isVirtual(true).build()));
+
+        // First run - executes child context
+        var result = runner.runUntilComplete("test");
+        assertEquals(ExecutionStatus.SUCCEEDED, result.getStatus());
+        assertEquals("result-test", result.getResult(String.class));
+        assertEquals(1, childExecutionCount.get());
+
+        // Second run - replays, should return cached result without re-executing
```
SilanHe, line 67: this comment is also misleading, the virtual child context was re-executed since `assertEquals(2, childExecutionCount.get(),...)` does not fail

sdk-integration-tests/src/test/java/software/amazon/lambda/durable/ChildContextIntegrationTest.java
```diff
@@ -72,6 +101,36 @@ void childContextExceptionPreservedOnReplay() {
...
+                            throw new IllegalArgumentException("bad input: " + input);
+                        },
+                        RunInChildContextConfig.builder().isVirtual(true).build()));
+
+        // First run - child context fails
+        var result = runner.run("test");
+        assertEquals(ExecutionStatus.FAILED, result.getStatus());
+        assertEquals(1, childExecutionCount.get());
+
+        // Second run - replays, should throw same exception without re-executing
```
SilanHe: repeat

sdk-integration-tests/src/test/java/software/amazon/lambda/durable/ChildContextIntegrationTest.java
SilanHe, line 67: nit: this is still the same

SilanHe (approved):

PR #368, opened 2026-05-01
move SuspendExecutionException to under Error

### LINKED ISSUE #369: [Feature]: avoid mis-handling of `SuspendExecutionException`
### What would you like?

Users might catch and swallow `SuspendExecutionException` by mistake if they catch `Exception` because `SuspendExecutionException` currently inherits `Exception`

An improvement can be done to move `SuspendExecutionException` from `Exception` to `Error`, so that users won't surprisingly receive `SuspendExecutionException` unless they catch `Throwable` or `Error`

### Possible Implementation

_No response_

No

### Does this require an RFC?

No

### Additional Context

_No response_

zhongkechen:

### Issue Link, if available

- Fixes #369 

### Description

When users catch `Exception` in their error handling code, they'll also catch `SuspendExecutionException` which needs to be rethrown to make durable function work.

The current workaround is to explicitly handle `SuspendExecutionException` before handling `Exception` like below
```
catch (SuspendExecutionException e) {
    throw e;
}
catch (Exception e) {
    // user's logic to handle exceptions
}
```

This change is to prevent `SuspendExecutionException` is mistakenly handled by users' exception handling logic.

With this change, the following exception handler code will not see `SuspendExecutionException` 
```
catch (Exception e) {
    // user's logic to handle exceptions
}
```

### Checklist

#### Unit Tests

#### Integration Tests

#### Examples

### DIFF
--- a/sdk-integration-tests/src/test/java/software/amazon/lambda/durable/CallbackIntegrationTest.java
+++ b/sdk-integration-tests/src/test/java/software/amazon/lambda/durable/CallbackIntegrationTest.java
@@ -271,10 +271,9 @@ class CallbackIntegrationTest {
                 ctx.waitForCallback("approval", String.class, (callbackId, stepCtx) -> {});
                 fail();
                 return "should not reach here";
+            } catch (SuspendExecutionException e) {
+                throw e;
             } catch (Exception e) {
-                if (e instanceof SuspendExecutionException) {
-                    throw e;
-                }
                 assertInstanceOf(CallbackFailedException.class, e);
                 throw e;
             }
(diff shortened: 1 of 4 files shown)


sdk-integration-tests/src/test/java/software/amazon/lambda/durable/CallbackIntegrationTest.java
```diff
@@ -271,10 +271,9 @@ void waitForCallbackCallbackFailed() {
                 ctx.waitForCallback("approval", String.class, (callbackId, stepCtx) -> {});
                 fail();
                 return "should not reach here";
+            } catch (SuspendExecutionException e) {
```
SilanHe, line 274: Do you still need to catch this?

zhongkechen: No, it wouldn't be necessary. I wanted to keep this to demonstrate how this exception should be caught and then rethrown if users are still using the older versions.

SilanHe: nit: I think we should remove from this commit since it's no longer applicable, right? I won't block on this though

SilanHe (approved):

PR #399, opened 2026-05-24
Retry interrupted at-most-once steps

### LINKED ISSUE #394: [Bug]: step does not retry with at_most_once_per_retry config
### Expected Behavior

When step is configured with at_most_once and retry, it should retry if the step failed without checkpoint (e.g. sandbox crashes).

### Actual Behavior

The step is not retried

### Steps to Reproduce

1. A step that failed at the first invocation without checkpoint failed, simulate lambda env crashes (e.g. System.exit(1)), succeeded on retry.
2. The step is configured with

```
StepConfig.builder()
                .semantics(StepSemantics.AT_MOST_ONCE_PER_RETRY)
                .retryStrategy((error, attempt) -> {
                    if (attempt >= 3) return RetryDecision.fail();
                    return RetryDecision.retry(Duration.ofSeconds(1));
                })
                .build()
```

3. The re-invocation should failed with `StepInterruptException` on first replay because we have `at most once` configured
4. It should be retried

### SDK Version

1.1.0

### Java Version

21

### Is this a regression?

No

### Last Working Version

_No response_

### Additional Context

_No response_

kiwigitops:
## Summary
- let interrupted `AT_MOST_ONCE_PER_RETRY` steps use the configured retry strategy instead of failing before retry handling
- keep the no-retry checkpoint-failure path covered and add a regression for retrying after a STARTED checkpoint replay

Closes #394.

## Testing
- `docker run --rm -v ${PWD}:/workspace -w /workspace maven:3.9-eclipse-temurin-21 mvn -pl sdk-integration-tests -am -Dtest=StepSemanticsIntegrationTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `git diff --check`

### DIFF
--- a/sdk/src/main/java/software/amazon/lambda/durable/operation/StepOperation.java
+++ b/sdk/src/main/java/software/amazon/lambda/durable/operation/StepOperation.java
@@ -165,10 +165,9 @@ public class StepOperation<T> extends SerializableDurableOperation<T> {
             errorObject = serializeException(exception);
         }
 
-        var isRetryable = !(exception instanceof StepInterruptedException);
         var retryDecision = config.retryStrategy().makeRetryDecision(exception, attempt);
 
-        if (isRetryable && retryDecision.shouldRetry()) {
+        if (retryDecision.shouldRetry()) {
             // Send RETRY
             var retryDelayInSeconds = Math.toIntExact(retryDecision.delay().toSeconds());
             var retryUpdate = OperationUpdate.builder()
--- a/sdk-integration-tests/src/test/java/software/amazon/lambda/durable/StepSemanticsIntegrationTest.java
+++ b/sdk-integration-tests/src/test/java/software/amazon/lambda/durable/StepSemanticsIntegrationTest.java
@@ -4,6 +4,7 @@ package software.amazon.lambda.durable;
 
 import static org.junit.jupiter.api.Assertions.*;
 
+import java.time.Duration;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.junit.jupiter.api.Test;
 import software.amazon.lambda.durable.config.StepConfig;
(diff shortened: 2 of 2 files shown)


sdk/src/main/java/software/amazon/lambda/durable/operation/StepOperation.java
```diff
@@ -165,10 +165,9 @@ private void handleStepFailure(Throwable exception, int attempt) {
             errorObject = serializeException(exception);
         }
 
-        var isRetryable = !(exception instanceof StepInterruptedException);
         var retryDecision = config.retryStrategy().makeRetryDecision(exception, attempt);
 
-        if (isRetryable && retryDecision.shouldRetry()) {
+        if (retryDecision.shouldRetry()) {
```
zhongkechen, line 170: will this change leading to retry and break the at-most-once semantic?

kiwigitops: No, this keeps the at-most-once guarantee scoped to each retry attempt.

For a replayed `STARTED` at-most-once step, we still do not call `executeStepLogic` for that same attempt. The interrupted attempt is converted through the retry strategy into either `RETRY`/`PENDING` or `FAIL`. User code only runs again after the operation becomes `READY`, which is the next retry attempt.

That is what the two tests are intended to pin down: `testAtMostOnceNoRetryFailsAfterCheckpointFailure` proves the started attempt is not re-executed when retries are disabled, and `testAtMostOnceRetriesAfterCheckpointFailure` proves a configured retry creates exactly one next-attempt execution.

zhongkechen: This would be a breaking change to users who rely on the existing behavior.

Issue #364, opened 2026-04-30
[Feature]: Durable Execution Event Hooks


zhongkechen:
### What would you like?

I would like to create a plugin which receive events from Durable Execution SDK when an event occurs, such as execution starts, operation starts, etc.

- onOperationStart, onOperationEnd, onOperationAttemptStart, onOperationAttemptEnd fire at most once per operation/attempt.
    - onOperationAttemptStart and onOperationAttemptEnd will manage a span that is a child of the span managed by onOperationStart and onOperationEnd. That is retry attempts for steps and wait-for-condition operations.
- onExecutionStart, onExecutionEnd fire at most once per execution
- onInvocationStart, onInvocationEnd fire at most once per invocation

proposed design: https://github.com/aws/aws-durable-execution-sdk-js/blob/otel-instrumentation-design-v2/packages/aws-durable-execution-sdk-js/src/documents/JAVA_SDK_OTEL_PLUGIN.md

### Possible Implementation

Related work in TypeScript SDK for reference:

* notes about plugin interface: https://github.com/aws/aws-durable-execution-sdk-js/blob/otel-instrumentation-design-v2/packages/aws-durable-execution-sdk-js/src/documents/PLUGIN_IMPLEMENTATION.md
* plugin interface: https://github.com/aws/aws-durable-execution-sdk-js/blob/otel-instrumentation-design-v2/packages/aws-durable-execution-sdk-js/src/types/plugin.ts
* To see how onOperationStart, onOperationEnd, onOperationAttemptStart, onOperationAttemptEnd are currently wired in the TS SDK, you'll have to check the [branch](https://github.com/aws/aws-durable-execution-sdk-js/tree/otel-instrumentation-design-v2). For example, [onOperationStart](https://github.com/aws/aws-durable-execution-sdk-js/blame/otel-instrumentation-design-v2/packages/aws-durable-execution-sdk-js/src/handlers/step-handler/step-handler.ts#L120).
* wiring of onExecutionStart, onInvocationStart, onExecutionEnd, onInvocationEnd: https://github.com/aws/aws-durable-execution-sdk-js/blob/otel-instrumentation-design-v2/packages/aws-durable-execution-sdk-js/src/with-durable-execution.ts
[...]

Issue #333, opened 2026-04-16
[Feature]:  OTel Integration


zhongkechen:
### What would you like?

Integrate with OTel

### Is this a breaking change?

No

### Does this require an RFC?

No

josh-green-awaze:
Events for operation start/replay and outcome would be very useful. Logging in general from the SDK could do with being a bit more verbose about when various things are happening, as I've found querying the execution logs truncates most Errors and manual logging wrapped around steps is required.

zhongkechen:
> Logging in general from the SDK could do with being a bit more verbose about when various things are happening

This can be achieved by switching the log level to `DEBUG` or `TRACE` in `log4j2.xml` as shown below.

https://github.com/aws/aws-durable-execution-sdk-java/blob/8712f6b36c3ec3ef294bc93d71836958bd5b4e72/examples/src/main/resources/log4j2.xml#L10

The error objects and the result payloads are intentionally left out because they're user data.

PR #403, opened 2026-05-25
feat: Use per-context replay status 

### LINKED ISSUE #383: [Feature]: Use per-context replay status for logging
### What would you like?

I would like to have accurate per-context replay status for logging.

Currently the replay status is global, which means it's inaccurate when there are multiple branches running concurrently. When one branch exits REPLAY mode, the whole execution will be in EXECUTION mode although some others branches are still replaying existing operations. This leads to inaccurate logging behavior as well as a blocker to correct plugin behavior (blocking #333 #364 )

### Possible Implementation

Implementation in JS for reference

https://github.com/aws/aws-durable-execution-sdk-js/blob/5f951b1bae4a83c24a8d40a024d0a1b5346677fb/packages/aws-durable-execution-sdk-js/src/context/durable-context/durable-context.ts#L231-L239

No

### Does this require an RFC?

No

### Additional Context

_No response_

ayushiahjolia:

### Issue Link, if available
Fixes https://github.com/aws/aws-durable-execution-sdk-java/issues/383

Adds per-context replay tracking to the Java SDK. This is a prerequisite for the plugin hooks feature (#364) and OTel integration (#333).

## Problem

Log suppression during replay used the global `ExecutionManager.isReplaying()` flag. When multiple branches run concurrently (parallel/map), the global flag flips to EXECUTION as soon as *any* branch encounters new work - causing logs in other branches that are still replaying to leak through.

The old per-context field (`BaseContextImpl.isReplaying`) was a one-shot transition guard used only by `BaseDurableOperation.execute()`. It was never used for log suppression and answered a different question ("has this context ever seen a non-cached op?") rather than ("is the next operation already checkpointed?").

## Solution

Added proactive per-context replay tracking in `DurableContextImpl`:

- **`replayMode`** field - initialized by checking if the next operation (via `peekNextOperationId()`) exists in checkpoint storage
- **`updateReplayStatus()`** - called after `nextOperationId()` in each operation method; peeks at the *following* operation and transitions out of replay if it doesn't exist
- **`isReplayingContext()`** - exposes the per-context replay status to the logger and future plugin hooks

This matches the Python SDK's `_track_replay()` pattern: after consuming the current operation's ID, peek at the next one - if it's not in storage, we're past the replay boundary.

### Checklist

#### Unit Tests

#### Integration Tests

#### Examples


sdk/src/test/java/software/amazon/lambda/durable/logging/DurableLoggerTest.java
```diff
@@ -42,7 +42,7 @@ void setUp() {
     }
 
     private DurableLogger createLogger(Mode mode, Suppression suppression) {
-        when(mockExecutionManager.isReplaying()).thenReturn(mode == Mode.REPLAYING);
+        when(mockExecutionManager.hasOperationsForContext(null)).thenReturn(mode == Mode.REPLAYING);
```
SilanHe: Sorry I'm a bit confused. So basically, we're equating

`Mode.REPLAYING` to `ExecutionManager.hasOperationsForContext(someContextId) = true`?

* https://github.com/aws/aws-durable-execution-sdk-java/blob/dbd11088f3ed977f8a24875e40f83ced55534562/sdk/src/main/java/software/amazon/lambda/durable/execution/ExecutionManager.java#L178

```
   /**
     * Checks whether there are any cached operations for the given parent context ID. Used to initialize per-context
     * replay state — a context starts in replay mode if the ExecutionManager has cached operations belonging to it.
     *
     * @param parentId the context ID to check (null for root context)
     * @return true if at least one operation exists with the given parentId
     */
    public boolean hasOperationsForContext(String parentId) {
        return operationStorage.values().stream().anyMatch(op -> Objects.equals(op.parentId(), parentId));
    }
```

* https://github.com/aws/aws-durable-execution-sdk-java/blame/dbd11088f3ed977f8a24875e40f83ced55534562/sdk/src/main/java/software/amazon/lambda/durable/context/BaseContextImpl.java#L42

Let's say we only partially completed the Context (1 of 3 steps within it are completed), wouldn't`ExecutionManager.hasOperationsForContext(someContextId)` still return true since the parent Id is present?

How does this isReplaying variable transition to false?

zhongkechen: I think the current implemtation of `Context.isReplaying` is not something we want. We can add a new method to be used by the plugin.

ayushiahjolia: Removed old isReplaying and replaced with new implementation - where we check next operation instead of current one.

sdk/src/main/java/software/amazon/lambda/durable/context/BaseContextImpl.java
```diff
@@ -97,16 +94,12 @@ public ExecutionManager getExecutionManager() {
...
     /**
-     * Transitions this context from replay to execution mode. Called when the first un-cached operation is encountered.
+     * Returns whether this context is currently in replay mode. The default implementation returns false. Subclasses
+     * that track per-context replay status (like DurableContextImpl) override this.
      */
-    public void setExecutionMode() {
-        this.isReplaying = false;
+    @Override
+    public boolean isReplayingContext() {
+        return false;
```
SilanHe, line 103: is this intended?

sdk/src/main/java/software/amazon/lambda/durable/context/BaseContextImpl.java
```diff
@@ -97,16 +94,12 @@ public ExecutionManager getExecutionManager() {
         return executionManager;
     }
 
-    /** Returns whether this context is currently in replay mode. */
-    @Override
-    public boolean isReplaying() {
```
zhongkechen, line 102: removing an existing method would be a breaking change. We'll need a major version upgrade if we really want to remove this.

sdk/src/main/java/software/amazon/lambda/durable/context/BaseContextImpl.java
ayushiahjolia: Yes, intentional. The default returns false (don't suppress) which is the safe default. Only DurableContextImpl overrides it with actual replay tracking. The only other subclass is StepContextImpl, whose user function only runs during real execution - so its logs should never be suppressed.

zhongkechen:
I think this change is no longer required by our plugin interface and the OTel plugin as we are going to switch to a backend based approach.

sdk/src/main/java/software/amazon/lambda/durable/context/BaseContextImpl.java
ayushiahjolia: Can mark it as deprecated and remove in future.

ayushiahjolia:
Closing this PR, this will be fixed by https://github.com/aws/aws-durable-execution-sdk-java/issues/435

Issue #396, opened 2026-05-20
[Feature]: Add SubType for


wangyb-A:
### What would you like?

I would like to add SubType for Step, Wait and Callback Operations

### Possible Implementation

Please refer to python and js SubType implementation:

- [JS SDK](https://github.com/aws/aws-durable-execution-sdk-js/blob/9ab7117becb4922b1e5d034ea12e2ff00cdf265c/packages/aws-durable-execution-sdk-js/src/types/core.ts#L133)

- [Python SDK](https://github.com/aws/aws-durable-execution-sdk-python/blob/168cee97aa01f9986d90a23a2e99e7e42bd9242b/src/aws_durable_execution_sdk_python/lambda_service.py#L94)

### Is this a breaking change?

No

### Does this require an RFC?

No

PR #405, opened 2026-05-25
feat(sdk): Add subtypes for Step, Wait, Callback and Invoke operations

nvasiu:

### Issue Link, if available

https://github.com/aws/aws-durable-execution-sdk-java/issues/396

### Description

Made sub-types mandatory for all operation types.
- Added new OperationSubType values for all operation types that were missing them (step, wait, callback and invoke). 
- Updated OperationSubType values to now include their corresponding parent OperationType.
- OperationSubType is now a required argument for OperationIdentifier. OperationType is no longer passed, instead the type value is retrieved from the OperationSubType.

Updated all locations that create operations using OperationIdentifier, to now pass OperationSubType.
- DurableContextImpl (creates step, wait, invoke, callback, childContext, map, parallel, and waitForCondition operations)
- ConcurrencyOperation (creates child map and parallel operations)
- All related tests

Additional Changes:

WaitForConditionOperation was inconsistent with other operations because it created it's own OperationIdentifier within it's constructor.
- For all other operations, DurableContextImpl creates the OperationIdentifier and then passes it to the operation constructors.
- Updated WaitForConditionOperation to work in the same way.

Updated BaseDurableOperation to remove null checks on sub-type. since the sub-type is now required.

ChildContextOperation explicitly checks for unreachable subtypes. Updated ChildContextOperation to check for the new subtype values.

Updated LocalMemoryExecutionClient.applyResult to preserve subtype when updating an operation. Previously the subtype value was not passed on to the updated operation, causing test failures.

### Checklist

#### Unit Tests

Have unit tests been written for these changes? Updated Tests

#### Integration Tests

Have integration tests been written for these changes? Updated Tests

#### Examples

### DIFF
--- a/sdk/src/main/java/software/amazon/lambda/durable/context/DurableContextImpl.java
+++ b/sdk/src/main/java/software/amazon/lambda/durable/context/DurableContextImpl.java
@@ -11,7 +11,6 @@ import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 import org.slf4j.LoggerFactory;
-import software.amazon.awssdk.services.lambda.model.OperationType;
 import software.amazon.lambda.durable.DurableCallbackFuture;
 import software.amazon.lambda.durable.DurableConfig;
 import software.amazon.lambda.durable.DurableContext;
@@ -146,7 +145,7 @@ public class DurableContextImpl extends BaseContextImpl implements DurableContex
 
         // Create and start step operation with TypeToken
         var operation = new StepOperation<>(
-                OperationIdentifier.of(operationId, name, OperationType.STEP), func, resultType, config, this);
+                OperationIdentifier.of(operationId, name, OperationSubType.STEP), func, resultType, config, this);
 
         operation.execute(); // Start the step (returns immediately)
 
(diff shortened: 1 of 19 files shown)


sdk/src/main/java/software/amazon/lambda/durable/context/DurableContextImpl.java
```diff
@@ -146,7 +146,11 @@ public <T> DurableFuture<T> stepAsync(
 
         // Create and start step operation with TypeToken
         var operation = new StepOperation<>(
-                OperationIdentifier.of(operationId, name, OperationType.STEP), func, resultType, config, this);
+                OperationIdentifier.of(operationId, name, OperationType.STEP, OperationSubType.STEP),
```
zhongkechen: We can update the implementation of `OperationIdentifier.of` to add a subtype if it's not provided, e.g.

```
    public static OperationIdentifier of(String operationId, String name, OperationType type) {
        return new OperationIdentifier(operationId, name, type, OperationSubType.valueOf(type.name()));
    }
```

zhongkechen (approved):

PR #408, opened 2026-05-27
feat(sdk): Add DurableExecutionPlugin interface and plugin runner

ayushiahjolia:

### Issue Link, if available
https://github.com/aws/aws-durable-execution-sdk-java/issues/364

### Description

Introduce the plugin system foundation for the durable execution SDK:
- DurableInstrumentationPlugin interface with lifecycle hooks for execution, invocation, operation, and attempt-level events
- PluginRunner that composes multiple plugins with fire-and-forget and callback-wrapping execution semantics
- Info records (InvocationInfo, OperationInfo, AttemptInfo, AttemptEndInfo, OperationEndInfo, ExecutionEndInfo, OperationChangeInfo)
- withPlugins(List<DurableInstrumentationPlugin>) config field in DurableConfig.Builder

### Checklist

#### Unit Tests

#### Integration Tests

#### Examples

### DIFF
--- a/sdk/src/main/java/software/amazon/lambda/durable/DurableConfig.java
+++ b/sdk/src/main/java/software/amazon/lambda/durable/DurableConfig.java
@@ -21,6 +21,7 @@ import software.amazon.awssdk.services.lambda.model.GetDurableExecutionStateRequ
 import software.amazon.lambda.durable.client.DurableExecutionClient;
 import software.amazon.lambda.durable.client.LambdaDurableFunctionsClient;
 import software.amazon.lambda.durable.logging.LoggerConfig;
+import software.amazon.lambda.durable.plugin.PluginRunner;
 import software.amazon.lambda.durable.retry.PollingStrategies;
 import software.amazon.lambda.durable.retry.PollingStrategy;
 import software.amazon.lambda.durable.serde.JacksonSerDes;
@@ -94,6 +95,7 @@ public final class DurableConfig {
     private final LoggerConfig loggerConfig;
     private final PollingStrategy pollingStrategy;
     private final Duration checkpointDelay;
+    private final PluginRunner pluginRunner;
 
     private DurableConfig(Builder builder) {
         this.durableExecutionClient = Objects.requireNonNullElseGet(
@@ -104,6 +106,7 @@ public final class DurableConfig {
         this.loggerConfig = Objects.requireNonNullElseGet(builder.loggerConfig, LoggerConfig::defaults);
         this.pollingStrategy = Objects.requireNonNullElse(builder.pollingStrategy, PollingStrategies.Presets.DEFAULT);
         this.checkpointDelay = Objects.requireNonNullElseGet(builder.checkpointDelay, () -> Duration.ofSeconds(0));
+        this.pluginRunner = PluginRunner.noOp();
 
         validateConfiguration();
     }
@@ -178,6 +181,20 @@ public final class DurableConfig {
      */
     public Duration getCheckpointDelay() {
         return checkpointDelay;
+    }
+
+    /**
+     * Gets the plugin runner that dispatches lifecycle events to registered plugins.
+     *
+     * <p>Currently returns a no-op runner. Plugin registration via config will be added when the plugin system is fully
+     * wired.
+     *
+     * @return PluginRunner instance (always no-op until plugin wiring is complete)
+     * @deprecated This is a preview API that is experimental and may be changed or removed in future releases.
+     */
+    @Deprecated
+    public PluginRunner getPluginRunner() {
+        return pluginRunner;
     }
 
     public void validateConfiguration() {
--- a/sdk/src/main/java/software/amazon/lambda/durable/plugin/DurableExecutionPlugin.java
+++ b/sdk/src/main/java/software/amazon/lambda/durable/plugin/DurableExecutionPlugin.java
@@ -1,0 +1,72 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package software.amazon.lambda.durable.plugin;
+
+/**
+ * Plugin interface for instrumenting durable execution lifecycle events.
+ *
+ * <p>Implement this interface to integrate observability tools (OpenTelemetry, Datadog, etc.) with the durable
+ * execution SDK. The SDK calls these hooks at key lifecycle points without requiring modifications to core SDK code.
+ *
+ * <p>All methods have default no-op implementations, allowing plugins to override only the hooks they need.
+ *
+ * <p>Plugin errors are isolated — exceptions thrown by plugin methods are caught and logged but never disrupt SDK
+ * execution.
+ *
+ * @deprecated This is a preview API that is experimental and may be changed or removed in future releases.
+ */
+@Deprecated
+public interface DurableExecutionPlugin {
+
+    // ─── Invocation-level hooks ──────────────────────────────────────────
+
+    /**
+     * Called at the start of each Lambda invocation. Use to set up per-invocation state (trace ID, invocation span).
+     *
+     * <p>Check {@link InvocationInfo#isFirstInvocation()} to detect the first invocation of an execution (useful for
+     * sampling decisions or execution-level span creation).
+     */
+    default void onInvocationStart(InvocationInfo info) {}
+
+    /**
+     * Called at the end of each Lambda invocation. Use to flush spans/metrics before Lambda freezes.
+     *
+     * <p>This hook is awaited — the SDK blocks until it returns. This is the only safe flush point before Lambda
+     * freezes the execution environment.
+     *
+     * <p>Check {@link InvocationEndInfo#invocationStatus()} to detect if the execution reached a terminal state in this
+     * invocation (useful for writing summary records or flushing final data).
+     */
+    default void onInvocationEnd(InvocationEndInfo info) {}
+
+    // ─── Operation-level hooks ───────────────────────────────────────────
+
+    /** Called when an operation starts (including replay). Use for logging/metrics that want replay visibility. */
+    default void onOperationStart(OperationInfo info) {}
+
+    /**
+     * Called when an operation reaches a terminal status for the first time (not on replay).
+     *
+     * <p>The OTel plugin creates operation spans here with backfilled start/end timestamps.
+     */
+    default void onOperationEnd(OperationEndInfo info) {}
+
+    // ─── User function hooks ─────────────────────────────────────────────
+
+    /**
+     * Called when a user-provided function starts executing. This fires for both step attempts (with {@code attempt}
+     * set) and child context functions (with {@code attempt} null).
+     *
+     * <p>This hook fires on the same thread as user code, so plugins can set OTel context via
+     * {@code Context.makeCurrent()} here.
+     */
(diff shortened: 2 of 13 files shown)


sdk/src/main/java/software/amazon/lambda/durable/plugin/AttemptEndInfo.java
```diff
@@ -0,0 +1,34 @@
...
+ * @param type operation type
+ * @param subType operation sub-type (may be null)
+ * @param parentId parent operation ID (null for root-level operations)
+ * @param startTimestamp when the attempt started
+ * @param endTimestamp when the attempt ended
+ * @param attempt 1-based attempt number
+ * @param outcome the attempt outcome (SUCCEEDED, FAILED, or RETRYING)
+ * @param error non-null if the attempt failed
+ * @param nextAttemptDelaySeconds non-null if outcome is RETRYING
+ * @experimental This record is experimental and may be changed or removed in future releases.
```
SilanHe: I'm not sure this tag actually exists in Javadoc, I think people use @deprecated instead with an explanation that it's experimental.

sdk/src/main/java/software/amazon/lambda/durable/plugin/PluginRunner.java
```diff
@@ -0,0 +1,179 @@
...
+
+        // Build the chain: innermost is fn, each plugin wraps the next
+        Supplier<T> chain = fn;
+        for (int i = plugins.size() - 1; i >= 0; i--) {
+            var plugin = plugins.get(i);
+            var next = chain;
+            chain = () -> {
+                try {
+                    return hookExtractor.apply(plugin).apply(next);
+                } catch (Exception e) {
```
SilanHe: I updated the JS runAsCallback as follows: https://github.com/aws/aws-durable-execution-sdk-js/blob/bcea797ed2ba70949b1b755f6b92064705ad2cb3/packages/aws-durable-execution-sdk-js/src/utils/plugin/plugin-runner.ts#L21

it will not swallow the errors in the customer provided function but it swallows the plugin exceptions.

sdk/src/main/java/software/amazon/lambda/durable/plugin/AttemptEndInfo.java
ayushiahjolia: `@deprecated` doesn't show the right intent. I'll use a clear Javadoc note instead - something like `<b>Preview API:</b> This API is experimental and may change in future releases.`

sdk/src/main/java/software/amazon/lambda/durable/plugin/PluginRunner.java
ayushiahjolia: Thanks for the pointer - updated runAsCallback to match. It now uses a guardedFn wrapper that captures user function errors and re-throws them after the chain completes, even if a plugin swallows them. Same behavior as the JS version.

sdk/src/main/java/software/amazon/lambda/durable/plugin/PluginRunner.java
```diff
@@ -0,0 +1,210 @@
...
+
+    // ─── Wrap hooks (chained, errors don't swallow user function) ────────
+
+    /**
+     * Chains wrap hooks from all plugins around the given function. Plugins are applied in order (first plugin is
+     * outermost wrapper). If a plugin's wrap hook throws, the SDK falls back to executing without that plugin's wrap.
+     *
+     * <p>Critically, if the user function (fn) throws, that error is always propagated — plugins cannot swallow it.
+     */
+    private <T> T runAsCallback(
```
zhongkechen: I don't think wrapping user function is necessary in Java. OTel Java SDK can set the context in thread local before calling user functions. JavaScript is doing this because OTel JS SDK doesn't provide a way of setting context in async local.

ayushiahjolia: Makes sense. I added wrap hooks for cross-SDK consistency and guaranteed scope closure, but the start/end pattern is more idiomatic for Java since OTel uses ThreadLocal-based context. I'll remove the wrap hooks.

sdk/src/main/java/software/amazon/lambda/durable/plugin/AttemptOutcome.java
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Possible outcomes for an operation attempt.
+ *
+ * <p><b>Preview API:</b> This API is experimental and may be changed or removed in future releases.
+ */
+public enum AttemptOutcome {
+    SUCCEEDED,
+    FAILED,
+    RETRYING
```
zhongkechen: PENDING? if `onOperationAttemptStart` is also called for contexts

sdk/src/main/java/software/amazon/lambda/durable/plugin/DurableInstrumentationPlugin.java
```diff
@@ -0,0 +1,70 @@
...
+ * execution SDK. The SDK calls these hooks at key lifecycle points without requiring modifications to core SDK code.
+ *
+ * <p>All methods have default no-op implementations, allowing plugins to override only the hooks they need.
+ *
+ * <p>Plugin errors are isolated — exceptions thrown by plugin methods are caught and logged but never disrupt SDK
+ * execution.
+ *
+ * <p><b>Preview API:</b> This API is experimental and may be changed or removed in future releases.
+ */
+public interface DurableInstrumentationPlugin {
```
zhongkechen: Is this `DurableInstrumentationPlugin` only one of more plugin types we're going to add?

sdk/src/main/java/software/amazon/lambda/durable/plugin/DurableInstrumentationPlugin.java
```diff
@@ -0,0 +1,70 @@
...
+
+    // ─── Operation Attempt-level hooks ───────────────────────────────────
+
+    /**
+     * Called before each attempt of a retryable operation (step, waitForCondition). Use to start an attempt span.
+     *
+     * <p>This hook fires on the same thread as user code, so plugins can set OTel context via
+     * {@code Context.makeCurrent()} here.
+     */
+    default void onOperationAttemptStart(AttemptInfo info) {}
```
zhongkechen: Will this also called for `CONTEXT` operations? If so, how do we get the `attempt` counts?

sdk/src/main/java/software/amazon/lambda/durable/plugin/ExecutionStatus.java
```diff
@@ -0,0 +1,13 @@
...
+package software.amazon.lambda.durable.plugin;
+
+/**
+ * Terminal status of a durable execution.
+ *
+ * <p><b>Preview API:</b> This API is experimental and may be changed or removed in future releases.
+ */
+public enum ExecutionStatus {
+    SUCCEEDED,
+    FAILED
```
zhongkechen: PENDING?

sdk/src/main/java/software/amazon/lambda/durable/plugin/DurableInstrumentationPlugin.java
ayushiahjolia: This is the single plugin interface - all plugins (OTel, metrics) implement it. Different plugins override different hooks based on what they need.

sdk/src/main/java/software/amazon/lambda/durable/plugin/DurableInstrumentationPlugin.java
ayushiahjolia: Updated - previously context operations were handled by separate wrap hooks (following JS pattern). Renamed to onUserFunctionStart/onUserFunctionEnd which now fires for both steps and contexts (following Python pattern). attempt is nullable - set for steps, null for contexts.

sdk/src/main/java/software/amazon/lambda/durable/plugin/AttemptOutcome.java
ayushiahjolia: No longer needed

sdk/src/main/java/software/amazon/lambda/durable/plugin/DurableInstrumentationPlugin.java
zhongkechen: The name `DurableInstrumentationPlugin` indicates this is for instrumentation only, is this true?

ayushiahjolia: Yes, right now it's observe-only, but that could change in the future. Renaming to DurableExecutionPlugin so the name doesn't box us into instrumentation-only semantics if we later add hooks with more control.

sdk/src/main/java/software/amazon/lambda/durable/plugin/DurableExecutionPlugin.java
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * <p>Implement this interface to integrate observability tools (OpenTelemetry, Datadog, etc.) with the durable
+ * execution SDK. The SDK calls these hooks at key lifecycle points without requiring modifications to core SDK code.
+ *
+ * <p>All methods have default no-op implementations, allowing plugins to override only the hooks they need.
+ *
+ * <p>Plugin errors are isolated — exceptions thrown by plugin methods are caught and logged but never disrupt SDK
+ * execution.
+ *
+ * <p><b>Preview API:</b> This API is experimental and may be changed or removed in future releases.
```
SilanHe: I think it's not that uncommon to use the @deprecated tag along with a message explaining it's a new preview feature. I think @deprecated tag has additional benefits of being formatted differently within an IDE compared to a custom message.

sdk/src/main/java/software/amazon/lambda/durable/plugin/PluginRunner.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public void onInvocationStart(InvocationInfo info) {
+        run(p -> p.onInvocationStart(info));
+    }
+
+    /**
+     * Called at the end of each invocation. Awaited — the SDK blocks until all plugins return, allowing plugins to
+     * flush spans/metrics before Lambda freezes.
+     */
+    public void onInvocationEnd(InvocationEndInfo info) {
```
SilanHe, line 64: are we certain we don't need the "wrap" methods for Java? seems like it would be much easier to implement the Java Otel plugin with the these methods than without them.

SilanHe: I won't block on this, we can always add them back later if we need them

sdk/src/main/java/software/amazon/lambda/durable/DurableConfig.java
```diff
@@ -383,6 +399,32 @@ public Builder withCheckpointDelay(Duration duration) {
...
+         * DurableConfig.builder()
+         *     .withPlugins(List.of(new OpenTelemetryDurablePlugin()))
+         *     .build();
+         * }</pre>
+         *
+         * @param plugins list of instrumentation plugins
+         * @return This builder
+         * @experimental This method is experimental and may be changed or removed in future releases.
+         */
+        public Builder withPlugins(List<DurableExecutionPlugin> plugins) {
```
zhongkechen: Let's not add this until we actually wire the plugins.

zhongkechen (approved):

PR #411, opened 2026-05-27
Update examples template to use pre-defined role

nvasiu:

### Issue Link, if available

N/A

### Description

Removed policies from examples SAM template and added a new RoleArn parameter. Now, instead of auto-generating a new role for each function, they will all now use the same pre-existing role

The new RoleArn parameter is passed in from a GitHub secret.

### Checklist

#### Unit Tests

Have unit tests been written for these changes? N/A

#### Integration Tests

#### Examples

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -68,13 +68,13 @@ jobs:
       - name: sam build
         run: | # add --no-cached  if debugging sam build
           sam build --debug --parameter-overrides \
-            'ParameterKey=Architecture,ParameterValue=x86_64 ParameterKey=JavaVersion,ParameterValue=java${{ matrix.java }}'
+            'ParameterKey=Architecture,ParameterValue=x86_64 ParameterKey=JavaVersion,ParameterValue=java${{ matrix.java }} ParameterKey=RoleArn,ParameterValue=${{ secrets.DURABLE_INTEGRATION_TEST_ROLE_ARN }}'
         working-directory: ./examples
       - name: sam deploy
         run: | 
           sam deploy --stack-name JavaSDKCloudBasedIntegrationTestStack-Java${{ matrix.java }}Runtime \
-            --resolve-image-repos --resolve-s3 --capabilities CAPABILITY_IAM --parameter-overrides \
-            'ParameterKey=Architecture,ParameterValue=x86_64 ParameterKey=JavaVersion,ParameterValue=java${{ matrix.java }}'
+            --resolve-image-repos --resolve-s3 --parameter-overrides \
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -68,13 +68,13 @@ jobs:
       - name: sam build
         run: | # add --no-cached  if debugging sam build
           sam build --debug --parameter-overrides \
-            'ParameterKey=Architecture,ParameterValue=x86_64 ParameterKey=JavaVersion,ParameterValue=java${{ matrix.java }}'
+            'ParameterKey=Architecture,ParameterValue=x86_64 ParameterKey=JavaVersion,ParameterValue=java${{ matrix.java }} ParameterKey=RoleArn,ParameterValue=${{ secrets.DURABLE_INTEGRATION_TEST_ROLE_ARN }}'
         working-directory: ./examples
       - name: sam deploy
         run: | 
           sam deploy --stack-name JavaSDKCloudBasedIntegrationTestStack-Java${{ matrix.java }}Runtime \
             --resolve-image-repos --resolve-s3 --capabilities CAPABILITY_IAM --parameter-overrides \
```
yaythomas: not necessary anymore?

`--capabilities CAPABILITY_IAM`

yaythomas (changes requested):

examples/template.yaml
```diff
@@ -31,6 +34,7 @@ Globals:
       ExecutionTimeout: 300
       RetentionPeriodInDays: 7
     Runtime: !Ref JavaVersion
+    Role: !Ref RoleArn
```
yaythomas: I don't think SAM supports this in Globals?

add Role: !Ref RoleArn to each individual function's Properties block instead of in Globals.

yaythomas (changes requested):

examples/template.yaml
```diff
@@ -62,18 +67,7 @@ Resources:
...
-              Action:
-                - lambda:CheckpointDurableExecutions
-                - lambda:GetDurableExecutionState
-                - lambda:InvokeFunction
-              Resource: !Sub "arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:simple-invoke-example-${JavaVersion}-runtime"
-            - Effect: Allow
-              Action:
-                - lambda:InvokeFunction
-              Resource: '*'
+      Role: !Ref RoleArn
```
zhongkechen, line 70: Does the role have the permission to invoke the other functions?

SilanHe: yeah make sure that the new role we create will only have invoke permissions on a subset of Lambda functions

.github/workflows/e2e-tests.yml
nvasiu: I'll try removing this, not sure if it will cause an error because of left over IAM resources

Update: Removed and tests passed

examples/template.yaml
nvasiu: Updating role

zhongkechen (approved):

SilanHe (approved):

yaythomas (approved):

PR #422, opened 2026-06-02
feat(plugin): Wire plugin hooks into SDK execution lifecycle

ayushiahjolia:

### Issue Link, if available
https://github.com/aws/aws-durable-execution-sdk-java/issues/364

### Description
Wire the DurableExecutionPlugin hooks into the SDK so registered plugins receive lifecycle events during execution

### Demo/Screenshots
Screenshot 2026-06-02 at 2 01 49 PM

### Checklist

#### Unit Tests

#### Integration Tests

#### Examples

### DIFF
--- a/sdk/src/main/java/software/amazon/lambda/durable/DurableConfig.java
+++ b/sdk/src/main/java/software/amazon/lambda/durable/DurableConfig.java
@@ -5,6 +5,8 @@ package software.amazon.lambda.durable;
 import java.io.IOException;
 import java.io.InputStream;
 import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Objects;
 import java.util.Properties;
 import java.util.concurrent.ExecutorService;
@@ -21,6 +23,7 @@ import software.amazon.awssdk.services.lambda.model.GetDurableExecutionStateRequ
 import software.amazon.lambda.durable.client.DurableExecutionClient;
 import software.amazon.lambda.durable.client.LambdaDurableFunctionsClient;
 import software.amazon.lambda.durable.logging.LoggerConfig;
+import software.amazon.lambda.durable.plugin.DurableExecutionPlugin;
 import software.amazon.lambda.durable.plugin.PluginRunner;
 import software.amazon.lambda.durable.retry.PollingStrategies;
 import software.amazon.lambda.durable.retry.PollingStrategy;
@@ -106,7 +109,7 @@ public final class DurableConfig {
         this.loggerConfig = Objects.requireNonNullElseGet(builder.loggerConfig, LoggerConfig::defaults);
         this.pollingStrategy = Objects.requireNonNullElse(builder.pollingStrategy, PollingStrategies.Presets.DEFAULT);
         this.checkpointDelay = Objects.requireNonNullElseGet(builder.checkpointDelay, () -> Duration.ofSeconds(0));
-        this.pluginRunner = PluginRunner.noOp();
+        this.pluginRunner = builder.plugins.isEmpty() ? PluginRunner.noOp() : new PluginRunner(builder.plugins);
 
         validateConfiguration();
     }
@@ -186,10 +189,9 @@ public final class DurableConfig {
     /**
      * Gets the plugin runner that dispatches lifecycle events to registered plugins.
      *
-     * <p>Currently returns a no-op runner. Plugin registration via config will be added when the plugin system is fully
-     * wired.
+     * <p>Returns a no-op runner if no plugins were registered via the builder.
      *
-     * @return PluginRunner instance (always no-op until plugin wiring is complete)
+     * @return PluginRunner instance (never null)
      * @deprecated This is a preview API that is experimental and may be changed or removed in future releases.
      */
     @Deprecated
@@ -291,6 +293,7 @@ public final class DurableConfig {
         private LoggerConfig loggerConfig;
         private PollingStrategy pollingStrategy;
         private Duration checkpointDelay;
+        private List<DurableExecutionPlugin> plugins = new ArrayList<>();
 
         public Builder() {}
 
@@ -397,6 +400,30 @@ public final class DurableConfig {
          */
         public Builder withCheckpointDelay(Duration duration) {
             this.checkpointDelay = duration;
+            return this;
+        }
+
+        /**
+         * Registers one or more plugins for lifecycle event instrumentation.
+         *
+         * <p>Plugins receive hooks at invocation, operation, and user function boundaries. Errors thrown by plugins are
+         * isolated and never disrupt SDK execution.
+         *
+         * <p>Calling this method replaces any previously registered plugins. Plugins are called in registration order.
+         *
+         * @param plugins the plugins to register
+         * @return This builder
+         * @throws NullPointerException if any plugin is null
+         * @deprecated This is a preview API that is experimental and may be changed or removed in future releases.
+         */
+        @Deprecated
+        public Builder withPlugins(DurableExecutionPlugin... plugins) {
+            Objects.requireNonNull(plugins, "Plugins array cannot be null");
+            var newPlugins = new ArrayList<DurableExecutionPlugin>(plugins.length);
+            for (var plugin : plugins) {
+                newPlugins.add(Objects.requireNonNull(plugin, "Plugin cannot be null"));
+            }
+            this.plugins = newPlugins;
             return this;
         }
 
--- a/sdk/src/main/java/software/amazon/lambda/durable/execution/DurableExecutor.java
+++ b/sdk/src/main/java/software/amazon/lambda/durable/execution/DurableExecutor.java
@@ -24,6 +24,10 @@ import software.amazon.lambda.durable.exception.IllegalDurableOperationException
 import software.amazon.lambda.durable.exception.UnrecoverableDurableExecutionException;
 import software.amazon.lambda.durable.model.DurableExecutionInput;
 import software.amazon.lambda.durable.model.DurableExecutionOutput;
+import software.amazon.lambda.durable.plugin.InvocationEndInfo;
+import software.amazon.lambda.durable.plugin.InvocationInfo;
+import software.amazon.lambda.durable.plugin.InvocationStatus;
+import software.amazon.lambda.durable.plugin.PluginRunner;
 import software.amazon.lambda.durable.serde.SerDes;
 import software.amazon.lambda.durable.util.ExceptionHelper;
 
@@ -48,11 +52,21 @@ public class DurableExecutor {
             TypeToken<I> inputType,
             BiFunction<I, DurableContext, O> handler,
             DurableConfig config) {
+        var pluginRunner = config.getPluginRunner();
         try (var executionManager = new ExecutionManager(input, config)) {
+            var isFirstInvocation = !executionManager.isReplaying();
+            var requestId = lambdaContext != null ? lambdaContext.getAwsRequestId() : null;
+            var executionArn = input.durableExecutionArn();
+
             executionManager.registerActiveThread(null);
             var handlerFuture = CompletableFuture.supplyAsync(
                     () -> {
                         executionManager.setCurrentThreadContext(new ThreadContext(null, ThreadType.CONTEXT));
+
+                        // onInvocationStart runs on the user thread so plugins can
+                        // inject ThreadLocal objects, update MDC, etc.
+                        pluginRunner.onInvocationStart(new InvocationInfo(requestId, executionArn, isFirstInvocation));
+
                         var userInput = extractUserInput(
                                 executionManager.getExecutionOperation(), config.getSerDes(), inputType);
                         // use try-with-resources to clear logger properties
@@ -76,6 +90,13 @@ public class DurableExecutor {
 
                                 // return PENDING if it's SuspendExecutionException
                                 if (cause instanceof SuspendExecutionException) {
+                                    fireOnInvocationEnd(
+                                            pluginRunner,
+                                            requestId,
+                                            executionArn,
+                                            isFirstInvocation,
+                                            InvocationStatus.PENDING,
+                                            null);
                                     return DurableExecutionOutput.pending();
                                 }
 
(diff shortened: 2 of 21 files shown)


sdk/src/main/java/software/amazon/lambda/durable/execution/DurableExecutor.java
```diff
@@ -48,7 +52,15 @@ public static <I, O> DurableExecutionOutput execute(
...
             BiFunction<I, DurableContext, O> handler,
             DurableConfig config) {
+        var pluginRunner = config.getPluginRunner();
         try (var executionManager = new ExecutionManager(input, config)) {
+            var isFirstInvocation = !executionManager.isReplaying();
+            var requestId = lambdaContext != null ? lambdaContext.getAwsRequestId() : null;
+            var executionArn = input.durableExecutionArn();
+
+            // Fire onInvocationStart plugin hook
+            pluginRunner.onInvocationStart(new InvocationInfo(requestId, executionArn, isFirstInvocation));
```
zhongkechen: I think we should move this call into user thread below, so that users can inject thread local objects, update MDC, etc.

sdk/src/main/java/software/amazon/lambda/durable/execution/DurableExecutor.java
```diff
@@ -85,17 +104,40 @@ public static <I, O> DurableExecutionOutput execute(
                                                 UnrecoverableDurableExecutionException
                                                         unrecoverableDurableExecutionException
                                         && unrecoverableDurableExecutionException.isRetryable()) {
+                                    fireOnInvocationEnd(
+                                            pluginRunner,
+                                            requestId,
+                                            executionArn,
+                                            isFirstInvocation,
+                                            InvocationStatus.FAILED,
```
zhongkechen: I think we need a different status in this case if the execution is going to be retried

sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java
```diff
@@ -132,11 +134,15 @@ public void execute() {
...
                 }
+                // Fire onOperationStart plugin hook (including replay)
+                fireOnOperationStart(existing);
                 replay(existing);
             } else {
                 if (durableContext.isReplaying()) {
                     this.durableContext.setExecutionMode();
                 }
+                // Fire onOperationStart plugin hook (first execution, no existing operation)
+                fireOnOperationStart(null);
```
zhongkechen, line 145: I'm afraid it doesn't work with `null`. The `OperationInfo` constructed would have a lot of null values.

sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java
```diff
@@ -245,13 +251,52 @@ protected Operation waitForOperationCompletion() {
...
+                    getName(),
+                    getType(),
+                    getSubType(),
+                    durableContext.getParentId(),
+                    null,
+                    durableContext.isReplaying(),
+                    attempt);
+            pluginRunner.onUserFunctionStart(userFunctionStartInfo);
+
+            boolean userFunctionEndFired = false;
```
zhongkechen: Why do we need this flag?

sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java
```diff
@@ -245,13 +251,52 @@ protected Operation waitForOperationCompletion() {
...
+
+            // Fire onUserFunctionStart on the user code thread
+            var userFunctionStartInfo = PluginInfoConverter.toUserFunctionStartInfo(
+                    getOperationId(),
+                    null,
+                    getName(),
+                    getType(),
+                    getSubType(),
+                    durableContext.getParentId(),
+                    null,
```
zhongkechen: `rawParentId`? we don't have such thing in Java SDK

sdk/src/main/java/software/amazon/lambda/durable/DurableConfig.java
```diff
@@ -400,6 +406,45 @@ public Builder withCheckpointDelay(Duration duration) {
...
+         * @param plugins the plugins to register
+         * @return This builder
+         * @throws NullPointerException if any plugin is null
+         * @deprecated This is a preview API that is experimental and may be changed or removed in future releases.
+         */
+        @Deprecated
+        public Builder withPlugins(DurableExecutionPlugin... plugins) {
+            Objects.requireNonNull(plugins, "Plugins array cannot be null");
+            for (var plugin : plugins) {
+                this.plugins.add(Objects.requireNonNull(plugin, "Plugin cannot be null"));
```
zhongkechen: Why this isn't replacing the existing plugins?

sdk/src/main/java/software/amazon/lambda/durable/DurableConfig.java
```diff
@@ -400,6 +406,45 @@ public Builder withCheckpointDelay(Duration duration) {
...
+         * <p>This is primarily used internally (e.g., by the test runner) to transfer an existing PluginRunner to a
+         * rebuilt DurableConfig without needing access to the original plugin list.
+         *
+         * @param pluginRunner the pre-built PluginRunner
+         * @return This builder
+         * @deprecated This is a preview API that is experimental and may be changed or removed in future releases.
+         */
+        @Deprecated
+        public Builder withPluginRunner(PluginRunner pluginRunner) {
+            this.pluginRunner = pluginRunner;
```
zhongkechen: are we going to allow users to implement their own `PluginRunner`? But `PluginRunner` isn't an interface

sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java
```diff
@@ -454,4 +504,41 @@ protected void validateReplay(Operation checkpointed) {
...
+    private PluginRunner getPluginRunner() {
+        var config = getContext().getDurableConfig();
+        return config != null ? config.getPluginRunner() : PluginRunner.noOp();
+    }
+
+    /** Fires onOperationStart plugin hook. */
+    private void fireOnOperationStart(Operation existing) {
+        var info = PluginInfoConverter.toOperationInfo(
+                existing,
+                getOperationId(),
```
zhongkechen: We can use `operationIdentifier` which contains operation id, type, subtype and name.

sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java
ayushiahjolia: flag is a minor defensive safety net to ensure onUserFunctionEnd is only fired once

sdk/src/main/java/software/amazon/lambda/durable/execution/DurableExecutor.java
ayushiahjolia: makes sense, I'll update it

sdk/src/main/java/software/amazon/lambda/durable/execution/DurableExecutor.java
ayushiahjolia: RETRYING?

sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java
ayushiahjolia: yes but we will add support for it in future, right? I can add an explicit comment that for now this will always be null

sdk/src/main/java/software/amazon/lambda/durable/DurableConfig.java
ayushiahjolia: whats the correct behavior? user needs to provide all plugins at once or incrementally?

sdk/src/main/java/software/amazon/lambda/durable/DurableConfig.java
ayushiahjolia: No, it's just for test, will mark it @VisibleForTesting, or should I go with different testing approach?

sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java
ayushiahjolia: good suggestion, will refactor

sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java
ayushiahjolia: After replacing individual arg with operationIdentifier, we should not have this issue. rawId/rawParentId and timestamps would still be null.

sdk/src/main/java/software/amazon/lambda/durable/DurableConfig.java
ayushiahjolia: Removed this and added getPlugins() to get a list.

sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java
SilanHe: what are you inferring Frank?

1. Use hashedParentId in Java since there's only the hashed ids in Java
2. or are you still looking to remove the idea of raw vs hashed in the interface even if we do not implement it yet. I can get behind not adding rawIds yet and adding them later but then we need to prepend the current ID fields with "hashed".

sdk/src/main/java/software/amazon/lambda/durable/DurableConfig.java
zhongkechen: It's less confusing if users provide all plugins at once

sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java
zhongkechen: Currently the value of OperationId in Java SDK is meaningless. We can change that to whatever format we like without breaking users. If we add raw ids to the interface, we'll have to stick to that specific raw ID format.

I think we can add something (either a raw id or something simpler for Java SDK like a sequence number under its parent context, or a global sequence id) later based on what is required for future features.

sdk/src/main/java/software/amazon/lambda/durable/execution/DurableExecutor.java
zhongkechen: Sounds good

sdk/src/main/java/software/amazon/lambda/durable/plugin/PluginInfoConverter.java
```diff
@@ -18,94 +17,46 @@ public final class PluginInfoConverter {
...
-                subType != null ? subType.getValue() : null,
+                identifier.operationId(),
+                null,
+                identifier.name(),
+                identifier.operationType() != null ? identifier.operationType().toString() : null,
+                identifier.subType() != null ? identifier.subType().getValue() : null,
                 parentId,
-                rawParentId,
+                null,
                 operation != null ? operation.startTimestamp() : null,
```
zhongkechen: It looks weird that a start event (an `onOperationStart` call) doesn't have the `startTimestamp`.

sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java
ayushiahjolia: Removed it

sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java
```diff
@@ -245,13 +251,41 @@ protected Operation waitForOperationCompletion() {
...
+    protected void runUserHandler(Runnable runnable, ThreadType threadType, Integer attempt) {
         String operationId = getOperationId();
         logger.debug("Starting user handler for operation {} ({})", operationId, threadType);
+        var pluginRunner = getPluginRunner();
         Runnable wrapped = () -> {
             executionManager.setCurrentThreadContext(new ThreadContext(operationId, threadType));
+
+            // Fire onUserFunctionStart on the user code thread
+            var userFunctionStartInfo = PluginInfoConverter.toUserFunctionStartInfo(
+                    operationIdentifier, durableContext.getParentId(), durableContext.isReplaying(), attempt);
```
zhongkechen, line 273: To avoid confusion, I replaced `isReplaying` with `isReplayingChildren` in the latest Python SDK change. And uses `attempt` to indicate whether it's the first time the user function is called: 1 for the first time and null afterwards as the attempt number is unknown for a CONTEXT operation.

sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java
```diff
@@ -132,11 +134,15 @@ public void execute() {
                 if (ExecutionManager.isTerminalStatus(existing.status())) {
                     replayCompletedOperation.set(true);
                 }
+                // Fire onOperationStart plugin hook (including replay)
+                fireOnOperationStart(existing);
```
zhongkechen, line 138: This is different from Python SDK where `onOperationStart` is called only once for each operation

sdk/src/main/java/software/amazon/lambda/durable/plugin/PluginInfoConverter.java
zhongkechen: And how does the plugin know if it's the first time the operation is started?

zhongkechen (approved):
Approved with some minor comments which can be addressed later.

sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java
ayushiahjolia: rawId is just a variable name, when we implement it we can tweak it to whatever we want. If we remove those fields, wouldn't it be a breaking change in future to add it?

maybe we can rename it to unhashedID?

ayushiahjolia: Discussed offline, removing this for now. We can add it back when we need it.

sdk/src/main/java/software/amazon/lambda/durable/plugin/PluginInfoConverter.java
ayushiahjolia: Using now() in startTimestamp for first operation call, and document it.

sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java
ayushiahjolia: Makes sense, too many replays! Will make similar change in Java.

sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java
ayushiahjolia: We will discuss this offline to align on correct behavior

Issue #385, opened 2026-05-13
[Feature]: Throw a user friendly error if the SDK is run in non-durable functions


zhongkechen:
### What would you like?

I would like to see a user friendly error message if this SDK is run in non-durable functions

Currently the exception and error message is very confusing in this case:
```
START RequestId: 38062828-caf8-4afb-87fa-f01d87cfbd76 Version: $LATEST
Cannot invoke "String.split(String, int)" because "this.durableExecutionArn" is null: java.lang.NullPointerException
java.lang.NullPointerException: Cannot invoke "String.split(String, int)" because "this.durableExecutionArn" is null
at software.amazon.lambda.durable.execution.ExecutionManager.<init>(ExecutionManager.java:86)
at software.amazon.lambda.durable.execution.DurableExecutor.execute(DurableExecutor.java:50)
at software.amazon.lambda.durable.DurableHandler.handleRequest(DurableHandler.java:133)
at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(Unknown Source)
at java.base/java.lang.reflect.Method.invoke(Unknown Source)
```

### Is this a breaking change?

No

### Does this require an RFC?

No

PR #424, opened 2026-06-04
feat(sdk): add user-friendly error for non-durable functions

nvasiu:

### Issue Link, if available

https://github.com/aws/aws-durable-execution-sdk-java/issues/385

### Description

- Added a check for when the SDK is run in a non-durable function, and throw a more user-friendly error message.
- Also added a test for this check.

Durable functions must have `DurableConfig` set up, and their invocation payloads will always include a `DurableExecutionArn` and `CheckpointToken`. If those values are missing, then the function does not have `DurableConfig` set up, so it is non-durable.

### Checklist

#### Unit Tests

Have unit tests been written for these changes? YES

#### Integration Tests

#### Examples

### DIFF
--- a/sdk/src/main/java/software/amazon/lambda/durable/DurableHandler.java
+++ b/sdk/src/main/java/software/amazon/lambda/durable/DurableHandler.java
@@ -130,6 +130,11 @@ public abstract class DurableHandler<I, O> implements RequestStreamHandler {
         var inputString = new String(inputStream.readAllBytes());
         logger.debug("Raw input from durable handler: {}", inputString);
         var input = serDes.deserialize(inputString, TypeToken.get(DurableExecutionInput.class));
+        // Durable function inputs must contain DurableExecutionArn and CheckpointToken
+        if (input.durableExecutionArn() == null || input.checkpointToken() == null) {
+            throw new IllegalStateException(
+                    "Unexpected payload provided to start the durable execution. DurableConfig must be set in Lambda function configuration.");
+        }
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/software/amazon/lambda/durable/DurableHandler.java
```diff
@@ -130,6 +130,10 @@ public final void handleRequest(InputStream inputStream, OutputStream outputStre
         var inputString = new String(inputStream.readAllBytes());
         logger.debug("Raw input from durable handler: {}", inputString);
         var input = serDes.deserialize(inputString, TypeToken.get(DurableExecutionInput.class));
+		// Durable function inputs must contain DurableExecutionArn, CheckpointToken and InitialExecutionState
+        if (input.durableExecutionArn() == null || input.checkpointToken() == null || input.initialExecutionState() == null) {
+            throw new IllegalStateException("Function not configured for durable execution. DurableConfig must be set in Lambda function configuration.");
```
zhongkechen: Interesting. Can't checkpointToken be `null` in the first invocation?

zhongkechen (commented):
Spotless failed

sdk/src/main/java/software/amazon/lambda/durable/DurableHandler.java
nvasiu: Checked our code: a checkpointToken is created from the first invocation, so this should never be null.

zhongkechen (approved):

Issue #435, opened 2026-06-08
Remaining work items for Java SDK OpenTelemetry plugin GA readiness


ayushiahjolia:
This ticket tracks the remaining open work items for the Java SDK OTel plugin before it can move from experimental/preview to GA. The plugin currently produces correct spans for steps, waits, retries, and child contexts using InMemorySpanExporter in tests, but several gaps remain for production readiness and correctness.

1. Replace SamplingUtil with TraceIdRatioBased sampler - Done (https://github.com/aws/aws-durable-execution-sdk-java/pull/436)
2. Add test validating spans in X-Ray / OTLP collector (not just InMemorySpanExporter) - Done (https://github.com/aws/aws-durable-execution-sdk-java/pull/444)
3. Update replay logic to use UpdatedOperationId list from backend - Done (https://github.com/aws/aws-durable-execution-sdk-java/pull/439)
4. Update virtual context behavior for Otel, onOperationStart replay behavior, onOperationStart possiblity during lambda suspension, Spans should be exported in each invocation where it was started/updated/completed - Done (https://github.com/aws/aws-durable-execution-sdk-java/pull/482)

Stretch goal -
(https://github.com/aws/aws-durable-execution-sdk-java/issues/479)
1. W3C traceparent extraction support 
2. No arg Otel plugin Constructor

Out of scope, but related -
1. Discuss correct behavior of logging during replay, when to switch execution mode? - https://github.com/aws/aws-durable-execution-sdk-java/issues/383

SilanHe:
Regarding 

> 5. Discuss/modify onOperationStart replay behavior

I think we can fire the onOperationStart hook when we see the STARTED status on the operation. Did you have something else in mind?

[UPDATE]: we are now including a isReplay field in the OperationInfo. This is false for the first time the onOperationStart is called for the operation and true for subsequent calls

SilanHe:
> 4. Discuss/Implement how to handle virtual context/operation

Sometimes a virtual context can span multiple invocations. So I think we should either trigger the hooks every time or never for virtual context.

I'm leaning towards triggering the hooks every time since it's representative of what's happening during the invocation. Virtual context have to be replayed every time. It's possible that the operations within it aren't (since their results are checkpointed), but technically we run the code within a virtual context every time since we don't checkpoint the result.
[...]

PR #439, opened 2026-06-09
feat(sdk): Add UpdatedOperationIds support for replay status tracking

ayushiahjolia:

### Issue Link, if available
https://github.com/aws/aws-durable-execution-sdk-java/issues/435

### Description
- Consume the new UpdatedOperationIds field from the backend invocation payload to determine which operations transitioned during suspension.
- Fire onOperationEnd plugin hook for replayed-completed operations that are in the updated set, enabling the OTel plugin to emit spans for operations that completed while Lambda was frozen.

### Demo/Screenshots
N/A

### Checklist

#### Unit Tests

#### Integration Tests

#### Examples

Has a new example been added for the change? (if applicable) I'll add tests validating actual spans/Otel behavior in separate PR.

### DIFF
--- a/sdk-testing/src/main/java/software/amazon/lambda/durable/testing/local/LocalMemoryExecutionClient.java
+++ b/sdk-testing/src/main/java/software/amazon/lambda/durable/testing/local/LocalMemoryExecutionClient.java
@@ -4,9 +4,11 @@ package software.amazon.lambda.durable.testing.local;
 
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -35,11 +37,12 @@ public class LocalMemoryExecutionClient implements DurableExecutionClient {
     private final EventProcessor eventProcessor = new EventProcessor();
     private final List<OperationUpdate> operationUpdates = new CopyOnWriteArrayList<>();
     private final Map<String, Operation> updatedOperations = new HashMap<>();
+    private final Set<String> operationIdsUpdatedSinceLastInvocation = new HashSet<>();
 
     @Override
     public CheckpointDurableExecutionResponse checkpoint(String arn, String token, List<OperationUpdate> updates) {
         operationUpdates.addAll(updates);
-        updates.forEach(this::applyUpdate);
+        updates.forEach(update -> applyUpdate(update, true));
 
         var newToken = UUID.randomUUID().toString();
 
@@ -116,6 +119,16 @@ public class LocalMemoryExecutionClient implements DurableExecutionClient {
         return existingOperations.values().stream().toList();
     }
 
+    /**
+     * Returns the list of operation IDs that have been updated since the last invocation, then clears the tracking set.
+     * This simulates the backend's {@code UpdatedOperationIds} field behavior.
+     */
+    public List<String> getUpdatedOperationIdsSinceLastInvocation() {
+        var ids = List.copyOf(operationIdsUpdatedSinceLastInvocation);
+        operationIdsUpdatedSinceLastInvocation.clear();
+        return ids;
+    }
+
     /** Build TestResult from current state. */
     public <O> TestResult<O> toTestResult(DurableExecutionOutput output, TypeToken<O> resultType, SerDes serDes) {
         var testOperations = existingOperations.values().stream()
@@ -139,7 +152,7 @@ public class LocalMemoryExecutionClient implements DurableExecutionClient {
             throw new IllegalStateException("Operation not found: " + stepName);
         }
         var startedOp = op.toBuilder().status(OperationStatus.STARTED).build();
-        updateOperation(null, startedOp);
+        updateOperation(null, startedOp, false);
     }
 
     /** Simulate fire-and-forget checkpoint loss by removing the operation entirely */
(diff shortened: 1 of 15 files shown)


sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java
```diff
@@ -136,6 +136,14 @@ public void execute() {
                 }
                 // Fire onOperationStart plugin hook (including replay)
                 fireOnOperationStart(existing);
```
zhongkechen, line 141: Why do we always fire onOperationStart when replay?

sdk/src/main/java/software/amazon/lambda/durable/execution/ExecutionManager.java
```diff
@@ -57,6 +57,7 @@ public class ExecutionManager implements AutoCloseable {
     private final String durableExecutionArn;
     private final AtomicReference<ExecutionMode> executionMode;
     private final DurableConfig durableConfig;
+    private final Set<String> updatedOperationIds;
```
zhongkechen: add `SinceLastInvocation` to make it more clear?

sdk-testing/src/main/java/software/amazon/lambda/durable/testing/local/LocalMemoryExecutionClient.java
```diff
@@ -238,5 +254,11 @@ private void updateOperation(OperationUpdate update, Operation op) {
         synchronized (updatedOperations) {
             updatedOperations.put(op.id(), op);
         }
+        // Only track operations updated outside of a checkpoint call (i.e., between invocations)
+        // for the updatedOperationIds field. Operations updated during a checkpoint are already
+        // visible to the SDK via the checkpoint response.
+        if (!withinCheckpoint) {
```
zhongkechen, line 257: Would be more explicit if add `withinCheckpoint` as a parameter

sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java
ayushiahjolia: That's existing behavior, I am going to fix it under https://github.com/aws/aws-durable-execution-sdk-java/issues/435

sdk/src/main/java/software/amazon/lambda/durable/execution/ExecutionManager.java
ayushiahjolia: I used the same name as DurableExecutionInput we get from backend, will update.

sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java
```diff
@@ -136,6 +136,14 @@ public void execute() {
...
+
+                // Fire onOperationEnd for operations that completed during suspension (between invocations).
+                // This enables the OTel plugin to emit spans for operations that transitioned while Lambda was frozen.
+                if (replayCompletedOperation.get()
+                        && executionManager.isOperationUpdatedSinceLastInvocation(getOperationId())) {
+                    fireOnOperationEnd(existing, null);
```
SilanHe: couldn't there be an error here? I'm thinking of an invoke operation which fails between invocations

ayushiahjolia: Yes, If an operation fails between invocations, the error won't propagate to the span - it'll look successful. I'll fix both this path and the existing onCheckpointComplete path (which has the same gap) to extract the ErrorObject from the operation's type-specific details and pass it as the error.

sdk/src/test/java/software/amazon/lambda/durable/execution/ExecutionManagerTest.java
```diff
@@ -139,4 +139,59 @@ void emptyInitialState() {
...
+                input,
+                DurableConfig.builder().withDurableExecutionClient(client).build());
+
+        assertFalse(manager.isOperationUpdatedSinceLastInvocation("1"));
+    }
+
+    @Test
+    void isOperationUpdatedSinceLastInvocation_handlesMultipleIds() {
+        client = TestUtils.createMockClient();
+        var initialState = CheckpointUpdatedExecutionState.builder()
```
SilanHe, line 184: nit: I'm surprised this doesn't throw even though 2,3,4 aren't in the initial state.

SilanHe (approved):

sdk/src/test/java/software/amazon/lambda/durable/execution/ExecutionManagerTest.java
ayushiahjolia: Let's discuss offline on expected behavior, if we should trust backend or validate operation ids.

sdk/src/main/java/software/amazon/lambda/durable/execution/ExecutionManager.java
```diff
@@ -109,6 +114,18 @@ public boolean isReplaying() {
         return executionMode.get() == ExecutionMode.REPLAY;
     }
 
+    /**
+     * Returns {@code true} if the given operation was updated since the last successful invocation. This is used by the
+     * OTel plugin to determine whether a replayed completed operation should emit a span — only operations that
```
SilanHe, line 119: just looking at this in retrospect, I think this is/will be used by the the SDK to determine whether to call onOperationStart and onOperationEnd hooks.

ayushiahjolia: I am not sure if there will be any onOperationStart during lambda suspension, but the purpose of this is for onOperationEnd for now.

sdk/src/test/java/software/amazon/lambda/durable/execution/ExecutionManagerTest.java
ayushiahjolia: We will rely on backend to provide valid operation ids and ignore unknown ones.

Issue #415, opened 2026-05-28
[Feature]: durable function handlers indirectly inherited from DurableHandler


zhongkechen:
### What would you like?

I would like to create durable function handlers that are indirectly inherited from `DurableHandler`, e.g.

```java
public abstract class AbstractDurableHandler<TRequest, TResponse> extends DurableHandler<DurableEnvelope, TResponse> {

  @Override
  public final TResponse handleRequest(final DurableEnvelope envelope, final DurableContext context) {
    //...
  }
}

public class InvoicesDurableHandler extends AbstractDurableHandler<GetInvoicesRequest, GetInvoicesResponse> {
    //...
}
```

### Is this a breaking change?

No

### Does this require an RFC?

No

PR #442, opened 2026-06-09
feat(sdk): support indirect DurableHandler inheritance

nvasiu:

### Issue Link, if available

https://github.com/aws/aws-durable-execution-sdk-java/issues/415

### Description

Added a `DurableHandler` constructor which accepts an input type explicitly.

This allows users to extend `DurableHandler` indirectly through an intermediate class, where the input type can't be inferred from the superclass.

The existing no-arg constructor still infers the input type from the superclass. This was left unchanged for backwards compatability.

### Checklist

#### Unit Tests

Have unit tests been written for these changes? YES

#### Integration Tests

#### Examples

### DIFF
--- a/sdk/src/main/java/software/amazon/lambda/durable/DurableHandler.java
+++ b/sdk/src/main/java/software/amazon/lambda/durable/DurableHandler.java
@@ -36,6 +36,17 @@ public abstract class DurableHandler<I, O> implements RequestStreamHandler {
     }
 
     /**
+     * Constructs a handler with an explicitly provided input type. Use this when the input type cannot be inferred from
+     * the generic superclass, such as when extending {@link DurableHandler} indirectly through an intermediate class.
+     *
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/software/amazon/lambda/durable/DurableHandler.java
```diff
@@ -35,6 +35,17 @@ protected DurableHandler() {
         this.config = createConfiguration();
```
zhongkechen, line 35: nit: we can let this default constructor calling the new one

zhongkechen (approved):

sdk/src/main/java/software/amazon/lambda/durable/DurableHandler.java
nvasiu: The default constructor uses `getClass()` right now, which doesn't work if I try to chain the constructors.

PR #464, opened 2026-06-17
refactor(otel): Add X-Ray context extraction with trace ID priority

ayushiahjolia:

### Issue Link, if available
https://github.com/aws/aws-durable-execution-sdk-java/issues/435

### Description
Refactors the OTel plugin's trace ID generation to use the X-Ray trace header propagated by the durable execution backend, replacing the external `opentelemetry-aws-xray-propagator` dependency with a lightweight hand-rolled parser.

The durable execution backend propagates the same `_X_AMZN_TRACE_ID` Root to all invocations of a given execution. By extracting and using this trace ID directly, all spans from all invocations naturally land in a single unified trace without needing ARN-derived hashing. This also enables parent span linkage so invocation spans appear as children of Lambda's X-Ray segment.

### Demo/Screenshots
N/A

### Checklist

#### Unit Tests

#### Integration Tests

Have integration tests been written for these changes? Yes, updated

#### Examples

Has a new example been added for the change? (if applicable) Will add Xray tests in following PR

### DIFF
--- a/otel-plugin/src/main/java/software/amazon/lambda/durable/otel/OpenTelemetryDurablePlugin.java
+++ b/otel-plugin/src/main/java/software/amazon/lambda/durable/otel/OpenTelemetryDurablePlugin.java
@@ -39,7 +39,22 @@ import software.amazon.lambda.durable.plugin.UserFunctionStartInfo;
  *   <li><b>Attempt span</b> — one per user function execution (step attempt, child context run)
  * </ul>
  *
- * <p>Uses deterministic span/trace IDs so all invocations of the same execution share a single trace.
+ * <p>Trace ID resolution:
+ *
+ * <ol>
+ *   <li>Uses the X-Ray trace ID from {@code _X_AMZN_TRACE_ID} when available. The durable execution backend propagates
+ *       the same Root to all invocations of the same execution, naturally unifying the trace.
+ *   <li>Falls back to a deterministic trace ID derived from the execution ARN (for local tests or non-Lambda
+ *       environments).
+ * </ol>
+ *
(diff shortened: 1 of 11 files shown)


otel-plugin/src/main/java/software/amazon/lambda/durable/otel/OpenTelemetryDurablePlugin.java
```diff
@@ -117,14 +139,36 @@ public OpenTelemetryDurablePlugin(
...
+        var extractedContext = contextExtractor.extract();
 
-        // Create invocation span as child of extracted context
+        if (extractedContext != null) {
+            // Use the X-Ray trace ID — backend propagates same Root across all invocations
+            idGenerator.setExtractedTraceId(extractedContext.traceId());
+        }
+        // If no extracted context, idGenerator falls back to ARN-derived trace ID
+
+        // Determine parent context for the invocation span
```
SilanHe, line 155: is this required? I don't think I'm creating an explicit parentSpan for the invocation span.

SilanHe (approved):

otel-plugin/src/main/java/software/amazon/lambda/durable/otel/OpenTelemetryDurablePlugin.java
ayushiahjolia: I think it will look better visually. 
```
├── Lambda segment (created by AWS automatically)
         │
         └── durable.invocation  
```
vs
```
├── Lambda segment (created by AWS automatically)
│
├── durable.invocation  
```
We can remove it later once we have demos ready and if we dont like it

PR #465, opened 2026-06-17
fix: align log metadata with other SDKs

### LINKED ISSUE #440: [fix]: Implement Option 3 in the field name discussion
### What would you like?

Implement Option 3 in https://github.com/aws/aws-durable-execution-sdk-java/discussions/438

Use the same field name `executionArn` in `DurableLogger` as the other SDKs do

### Possible Implementation

_No response_

No

### Does this require an RFC?

No

### Additional Context

_No response_

zhongkechen:

### Issue Link, if available

Fixes #440

Partially #383

### Description

Align durable execution log metadata with the JavaScript and Python SDKs by moving MDC context management into `DurableLogger` and updating the emitted key names. This also separates logger scope cleanup from durable context objects so context lifecycles are clearer and less error-prone.

- Align MDC field names with the other SDKs by default (fixes #440)
  - `executionArn` instead of `durableExecutionArn`
  - root/child durable contexts emit `operationId` / `operationName`
  - step contexts continue to emit `operationId`, `operationName`, and `attempt`
  - Add `LoggerConfig.oldKeyNames` to preserve backward compatibility for consumers that still expect the previous MDC field names.

- Switch `DurableLogger` to use per context replay status (partially #383)
    - fixed a bug in per context replay status that causes the status to be incorrect initially
    - The value of the replay status in context hasn't been updated to use new `updatedOperationIds` field yet. We'll follow up on this when the new field is ready.

- Refactor `DurableLogger` to manage MDC state through attach/detach helpers instead of storing a logger per durable context, allowing users to use the same logger in anywhere and emit different metadata relying on the context.
    - Remove `AutoCloseable` / `close()` responsibilities from durable context types and make logger scoping explicit at user-code execution boundaries.
    - Update `DurableExecutor`, `StepOperation`, `ChildContextOperation`, and `WaitForConditionOperation` to attach/detach logger scope around user code.
    - `getLogger()` can wrap a user `Logger`, allowing users to customize the logger, e.g. `log.logger` field in metadata

### Demo/Screenshots

image

### Checklist

### Testing

Added a new case `testLoggingExample` to run `logginer-example`.

Expanded `DurableLoggerTest` coverage for:

- replay log suppression behavior
- MDC population for durable and step contexts
- MDC cleanup on detach
- replay-to-execution transition behavior
- all supported log levels
- null request ID handling
- old vs new MDC key expectations

#### Unit Tests

Yes

#### Integration Tests

No

#### Examples

N/A

### DIFF
--- a/docs/advanced/error-handling.md
+++ b/docs/advanced/error-handling.md
@@ -3,13 +3,13 @@
 The SDK throws specific exceptions to help you handle different failure scenarios:
 
 ```
+Error
+└── DurableExecutionError                 - Internal SDK control-flow/error base type
+    └── SuspendExecutionException         - Internal signal used by the SDK to suspend execution
+                                           (for example during `wait()`, `waitForCallback()`, and
+                                           `waitForCondition()`). User code should not catch it.
+
 RuntimeException
-├── SuspendExecutionException          - Internal control-flow exception thrown by the SDK to suspend execution
-│                                        (e.g., during wait(), waitForCallback(), waitForCondition()).
-│                                        The SDK catches this internally — you will never see it unless you have
-│                                        a broad catch(Exception) block around durable operations. If caught
-│                                        accidentally, you MUST re-throw it so the SDK can suspend correctly.
-│
 └── DurableExecutionException              - General durable exception
     ├── SerDesException                    - Serialization and deserialization exception.
     ├── UnrecoverableDurableExecutionException - Execution cannot be recovered. The durable execution will be immediately terminated.
@@ -52,7 +52,11 @@ try {
 
 ### Handling SuspendExecutionException
 
-If you have a broad `catch (Exception e)` block around durable operations, you must re-throw `SuspendExecutionException` to let the SDK suspend correctly:
+`SuspendExecutionException` is an internal SDK control-flow signal. It extends `Error`, not `Exception`, so a
+normal `catch (Exception e)` block will not intercept it.
+
+The real risk is code that catches `Throwable`, or code that explicitly catches `SuspendExecutionException`. In those
+cases, you must re-throw it immediately so the SDK can suspend the execution correctly.
 
 ```java
 try {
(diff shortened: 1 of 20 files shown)


docs/advanced/error-handling.md
```diff
@@ -52,16 +52,36 @@ try {
 
 ### Handling SuspendExecutionException
 
-If you have a broad `catch (Exception e)` block around durable operations, you must re-throw `SuspendExecutionException` to let the SDK suspend correctly:
+`SuspendExecutionException` is an internal SDK control-flow signal. It extends `Error`, not `Exception`, so a
```
SilanHe, line 55: and you're also updating these error handling docs? is this intended?

sdk/src/main/java/software/amazon/lambda/durable/operation/StepOperation.java
```diff
@@ -104,7 +106,10 @@ private void executeStepLogic(int attempt) {
             // use a try-with-resources to
             // - add thread id/type to thread local when the step starts
             // - clear logger properties when the step finishes
-            try (StepContext stepContext = getContext().createStepContext(getOperationId(), getName(), attempt)) {
+            StepContext stepContext = getContext().createStepContext(getOperationId(), getName(), attempt);
+            BaseContextImpl.setCurrentContext(stepContext);
+
+            try (var ignored = DurableLogger.attachContext()) {
```
SilanHe, line 112: I guess it makes sense that you can use the logger within a step? I didn't know it would be the case since I considered it to be on a similar level to other operations and thought they shouldn't be nested

SilanHe (approved):

docs/advanced/error-handling.md
zhongkechen: Yeah, obsolete docs. Should be updated in this PR or a separate one

sdk/src/main/java/software/amazon/lambda/durable/operation/StepOperation.java
zhongkechen: Previously, it was not supposed to use a logger from another context. This restriction is removed with this change. Any logger from our `getLogger()` method is fully context aware:
- The same logger behaves differently in different context. 
- And also different loggers from different `getLogger()` calls behave the same in the same context.

ayushiahjolia (approved):

PR #480, opened 2026-06-19
fix: operations always return deserialized result

### LINKED ISSUE #456: [Bug]: Child operation should call ser & des before return the result
### Expected Behavior

I would like the the child operatoin to return the deseralized of seralized result on first run. So that the first run and replay will have same result.

Note that the ser/des should be applied to all child calls, including normal, large payload, virtual context.

### Actual Behavior

Only normal child operation will ser / des the result. We should add it for large payload and virtual context senario.

### Steps to Reproduce

n/a

### SDK Version

1.2.1

### Java Version

17

### Is this a regression?

No

### Last Working Version

_No response_

### Additional Context

_No response_

zhongkechen:

### Issue Link, if available

Fix #456 

### Description

Call deserialize immediately after serialize to make sure the serialized result is always deserialize-able

### Checklist

#### Unit Tests

#### Integration Tests

#### Examples

### DIFF
--- a/sdk/src/main/java/software/amazon/lambda/durable/operation/ChildContextOperation.java
+++ b/sdk/src/main/java/software/amazon/lambda/durable/operation/ChildContextOperation.java
@@ -149,6 +149,8 @@ public class ChildContextOperation<T> extends SerializableDurableOperation<T> {
     }
 
     private void handleChildContextSuccess(T result) {
+        var serialized = serializeResult(result);
+
         if (replayChildren.get() || isVirtual || parentOperation != null && parentOperation.isOperationCompleted()) {
             // Skip checkpointing if
             // - parent ConcurrencyOperation has already completed, preventing race conditions where a child finishes
@@ -162,13 +164,11 @@ public class ChildContextOperation<T> extends SerializableDurableOperation<T> {
             }
             markAlreadyCompleted();
         } else {
-            checkpointSuccess(result);
+            checkpointSuccess(result, serialized);
         }
     }
 
-    private void checkpointSuccess(T result) {
-        var serialized = serializeResult(result);
-
+    private void checkpointSuccess(T result, String serialized) {
         if (serialized == null || serialized.getBytes(StandardCharsets.UTF_8).length < LARGE_RESULT_THRESHOLD) {
             sendOperationUpdate(
                     OperationUpdate.builder().action(OperationAction.SUCCEED).payload(serialized));
--- a/sdk/src/main/java/software/amazon/lambda/durable/operation/ChildContextOperation.java
+++ b/sdk/src/main/java/software/amazon/lambda/durable/operation/ChildContextOperation.java
@@ -149,7 +149,7 @@ public class ChildContextOperation<T> extends SerializableDurableOperation<T> {
     }
 
     private void handleChildContextSuccess(T result) {
-        var serialized = serializeResult(result);
+        var serializedResult = serializeAndDeserializeResult(result);
 
         if (replayChildren.get() || isVirtual || parentOperation != null && parentOperation.isOperationCompleted()) {
             // Skip checkpointing if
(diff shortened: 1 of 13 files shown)


sdk/src/main/java/software/amazon/lambda/durable/operation/SerializableDurableOperation.java
```diff
@@ -101,7 +101,11 @@ protected T deserializeResult(String result) {
      * @return the serialized string
      */
     protected String serializeResult(T result) {
-        return resultSerDes.serialize(result);
+        var serialized = resultSerDes.serialize(result);
+        if (shouldValidateSerializationRoundTrip()) {
+            deserializeResult(serialized);
```
wangyb-A: will it call the deserialize twice? i.e. if they write to s3 during deserialization, the context will be write twice?

ayushiahjolia (approved):

sdk/src/main/java/software/amazon/lambda/durable/operation/StepOperation.java
```diff
@@ -144,9 +144,11 @@ private void checkpointStarted() {
     }
 
     private void handleStepSucceeded(T result) {
+        var serializedResult = serializeAndDeserializeResult(result);
```
ayushiahjolia, line 147: nit: `serializedResult` is a bit confusing, maybe parsedResult is better?

sdk/src/main/java/software/amazon/lambda/durable/operation/StepOperation.java
```diff
@@ -144,9 +144,11 @@ private void checkpointStarted() {
...
+
         // Send SUCCEED
         var successUpdate =
-                OperationUpdate.builder().action(OperationAction.SUCCEED).payload(serializeResult(result));
+                OperationUpdate.builder().action(OperationAction.SUCCEED).payload(serializedResult.serialized());
```
ayushiahjolia, line 151: Are we using serialized here because we re-execute step during replay? why can't we use raw then?

sdk/src/main/java/software/amazon/lambda/durable/operation/ChildContextOperation.java
```diff
@@ -149,26 +149,26 @@ private void executeChildContext() {
...
         } else {
-            checkpointSuccess(result);
+            checkpointSuccess(serializedResult.deserialized(), serializedResult.serialized());
         }
     }
 
-    private void checkpointSuccess(T result) {
-        var serialized = serializeResult(result);
-
+    private void checkpointSuccess(T result, String serialized) {
```
wangyb-A, line 171: `T result` is this required?

wangyb-A (approved):

sdk/src/main/java/software/amazon/lambda/durable/operation/StepOperation.java
zhongkechen: `payload` is a string field

sdk/src/main/java/software/amazon/lambda/durable/operation/ChildContextOperation.java
zhongkechen: yes, when caching large payload

sdk/src/main/java/software/amazon/lambda/durable/operation/StepOperation.java
zhongkechen: any suggestions?

ayushiahjolia: `parsedResult`?

sdk/src/main/java/software/amazon/lambda/durable/operation/SerializableDurableOperation.java
zhongkechen: Yes, `deserialize` will be called again each time users call `get()`. We can improve this by having a cached value for `get()`

Issue #507, opened 2026-06-30
OTel plugin produces disconnected X-Ray traces and groups retry attempts under wrong invocation


ayushiahjolia:
The OTel plugin has three issues that degrade the X-Ray tracing experience:
1. When a step retries across invocations, both attempts appear under the second invocation because the operation spans share the same deterministic span ID.
2. Durable execution spans land in a separate X-Ray trace from the Lambda segments because System.getenv("_X_AMZN_TRACE_ID") returns stale values in Java Lambda runtimes.
3. Durable spans don't appear as a separate service node in the trace map (unlike the JS SDK), making them harder to identify in complex traces.

PR #508, opened 2026-06-30
fix(otel): Fix X-Ray trace integration and cross-invocation retry spans

ayushiahjolia:

### Issue Link, if available
https://github.com/aws/aws-durable-execution-sdk-java/issues/507

### Description

Fixes OTel plugin X-Ray integration: cross-invocation retry span grouping, trace context extraction, and trace map display to match the JS SDK behavior.

1. **Retry attempts grouped under wrong invocation** - When a step failed in invocation 1 and retried in invocation 2, both operation spans had the same deterministic span ID, causing trace viewers to merge them.
2. **Durable spans in separate trace** — `System.getenv("_X_AMZN_TRACE_ID")` returned stale/null values due to JVM env caching, so the plugin fell back to an ARN-derived trace ID disconnected from Lambda's X-Ray segments.
3. **Spans nested inside Lambda function instead of separate node** - Missing`service.name` resource and `SpanKind.SERVER` prevented X-Ray from creating a distinct service node for durable execution spans.

### Demo/Screenshots
Screenshot 2026-06-30 at 12 33 02 PM

### Checklist

#### Unit Tests

#### Integration Tests

#### Examples

Has a new example been added for the change? (if applicable) Updated existing ones.

### DIFF
--- a/examples/template.yaml
+++ b/examples/template.yaml
@@ -370,9 +370,6 @@ Resources:
         - !Sub
           - arn:aws:lambda:${AWS::Region}:901920570463:layer:aws-otel-java-agent-${AdotArch}-ver-1-32-0:6
           - AdotArch: amd64
-      Environment:
-        Variables:
-          AWS_LAMBDA_EXEC_WRAPPER: /opt/otel-handler
 
   OtelXRayWaitExampleFunction:
     Type: AWS::Serverless::Function
@@ -389,9 +386,6 @@ Resources:
         - !Sub
           - arn:aws:lambda:${AWS::Region}:901920570463:layer:aws-otel-java-agent-${AdotArch}-ver-1-32-0:6
           - AdotArch: amd64
-      Environment:
-        Variables:
-          AWS_LAMBDA_EXEC_WRAPPER: /opt/otel-handler
 
   RetryInvokeExampleFunction:
     Type: AWS::Serverless::Function
--- a/otel-plugin/src/main/java/software/amazon/lambda/durable/otel/OtelPlugin.java
+++ b/otel-plugin/src/main/java/software/amazon/lambda/durable/otel/OtelPlugin.java
@@ -5,16 +5,20 @@ package software.amazon.lambda.durable.otel;
 import static software.amazon.lambda.durable.otel.SpanAttributes.*;
 
 import io.opentelemetry.api.common.AttributeKey;
+import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.api.trace.SpanContext;
+import io.opentelemetry.api.trace.SpanKind;
 import io.opentelemetry.api.trace.StatusCode;
 import io.opentelemetry.api.trace.TraceFlags;
 import io.opentelemetry.api.trace.TraceState;
 import io.opentelemetry.api.trace.Tracer;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.Scope;
+import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.sdk.trace.SdkTracerProvider;
 import io.opentelemetry.sdk.trace.SdkTracerProviderBuilder;
+import io.opentelemetry.semconv.ServiceAttributes;
 import java.time.Instant;
 import java.util.concurrent.ConcurrentHashMap;
 import org.slf4j.Logger;
@@ -128,6 +132,12 @@ public class OtelPlugin implements DurableExecutionPlugin {
     public OtelPlugin(
             SdkTracerProviderBuilder tracerProviderBuilder, ContextExtractor contextExtractor, boolean enableMdc) {
         this.idGenerator = new DeterministicIdGenerator();
+
+        // Set service.name to "invocation" — X-Ray uses this as the display name for SERVER spans,
+        // creating a separate service node in the trace map labeled "invocation".
+        var resource = Resource.create(Attributes.of(ServiceAttributes.SERVICE_NAME, "invocation"));
+        tracerProviderBuilder.addResource(resource);
+
         this.tracerProvider = tracerProviderBuilder.setIdGenerator(idGenerator).build();
         this.tracer = tracerProvider.get(INSTRUMENTATION_NAME);
         this.contextExtractor = contextExtractor;
@@ -152,10 +162,11 @@ public class OtelPlugin implements DurableExecutionPlugin {
         }
         // If no extracted context, idGenerator falls back to ARN-derived trace ID
 
-        // Determine parent context for the invocation span
+        // Determine parent context for the invocation span.
         Context parentContext;
         if (extractedContext != null && extractedContext.parentSpanId() != null) {
-            // Create a remote parent span context from the X-Ray Parent field
+            // X-Ray header has parent — create the invocation span as a child of that segment.
+            // This connects our OTLP-exported spans to the Lambda service's X-Ray segments.
             var parentSpanContext = SpanContext.createFromRemoteParent(
                     extractedContext.traceId(),
                     extractedContext.parentSpanId(),
@@ -166,8 +177,10 @@ public class OtelPlugin implements DurableExecutionPlugin {
             parentContext = Context.root();
         }
 
-        // Create invocation span as child of Lambda's X-Ray segment (via Parent field)
-        var spanBuilder = tracer.spanBuilder("durable.invocation")
+        // Create a SERVER span to establish a separate X-Ray service node.
+        // X-Ray uses service.name for the segment display name.
+        var spanBuilder = tracer.spanBuilder("invocation")
+                .setSpanKind(SpanKind.SERVER)
                 .setParent(parentContext)
                 .setAttribute(DURABLE_EXECUTION_ARN, info.durableExecutionArn())
                 .setAttribute(DURABLE_FIRST_INVOCATION, info.isFirstInvocation());
@@ -227,8 +240,6 @@ public class OtelPlugin implements DurableExecutionPlugin {
     public void onOperationStart(OperationInfo info) {
         if (info.id() == null) return;
 
-        idGenerator.setNextSpanOperationId(info.id());
-
         var parentContext = resolveParentContext(info.parentId());
 
         var spanBuilder = tracer.spanBuilder(spanName(info.type(), info.subType(), info.name()))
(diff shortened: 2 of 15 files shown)


examples/template.yaml
```diff
@@ -370,9 +370,6 @@ Resources:
         - !Sub
           - arn:aws:lambda:${AWS::Region}:901920570463:layer:aws-otel-java-agent-${AdotArch}-ver-1-32-0:6
           - AdotArch: amd64
-      Environment:
-        Variables:
-          AWS_LAMBDA_EXEC_WRAPPER: /opt/otel-handler
```
SilanHe, line 375: how come we don't need this anymore?

sdk/src/main/java/software/amazon/lambda/durable/plugin/OperationInfo.java
```diff
@@ -27,4 +30,5 @@ public record OperationInfo(
         String subType,
         String parentId,
         Instant startTimestamp,
-        Instant endTimestamp) {}
+        Instant endTimestamp,
+        boolean isContinuation) {}
```
SilanHe: nit: we're using isReplay in JS (I think it can still be confusing but replay is a terminology that is more commonly used around durable functions). I can be convinced otherwise

SilanHe: how is this new variable currently being used? I'm not seeing changes in the sdk itself

zhongkechen: `Replay` is applicable only to user functions in Context operations. I think we lack a term for the other operations.

otel-plugin/src/main/java/software/amazon/lambda/durable/otel/OtelPlugin.java
```diff
@@ -152,10 +161,17 @@ public void onInvocationStart(InvocationInfo info) {
         }
         // If no extracted context, idGenerator falls back to ARN-derived trace ID
```
zhongkechen, line 163: Is idGenerator using the trace id from previous invocation?

otel-plugin/src/main/java/software/amazon/lambda/durable/otel/OtelPlugin.java
```diff
@@ -128,6 +131,12 @@ public OtelPlugin(SdkTracerProviderBuilder tracerProviderBuilder, ContextExtract
     public OtelPlugin(
             SdkTracerProviderBuilder tracerProviderBuilder, ContextExtractor contextExtractor, boolean enableMdc) {
         this.idGenerator = new DeterministicIdGenerator();
+
+        // Set service.name to "invocation" — X-Ray uses this as the display name for SERVER spans,
+        // creating a separate service node in the trace map labeled "invocation".
+        var resource = Resource.create(Attributes.of(AttributeKey.stringKey("service.name"), "invocation"));
```
zhongkechen: nit: use SERVICE_NAME constant from opentelemetry package.

otel-plugin/src/main/java/software/amazon/lambda/durable/otel/OtelPlugin.java
```diff
@@ -128,6 +131,12 @@ public OtelPlugin(SdkTracerProviderBuilder tracerProviderBuilder, ContextExtract
...
+        tracerProviderBuilder.setResource(resource);
```
zhongkechen: This overrides what users have configured for the resource in `tracerProviderBuilder`. Consider merge the resources and override only service name.

otel-plugin/src/main/java/software/amazon/lambda/durable/otel/OtelPlugin.java
ayushiahjolia: yes, it uses the same X-Ray Root trace ID by backend for every invocation of the same execution. If X-Ray context isn't available, it falls back to a deterministic hash of the execution ARN, which also produces the same value across invocations.

sdk/src/main/java/software/amazon/lambda/durable/plugin/OperationInfo.java
ayushiahjolia: > how is this new variable currently being used? I'm not seeing changes in the sdk itself 

New variable is used in https://github.com/aws/aws-durable-execution-sdk-java/pull/508/changes#diff-2cc56b0a18e2f9c269690a7b187c6db73771ab363612a14dbd7749f38bce951dR256 in Otel Plugin.

The SDK core sets the flag based on whether the operation already existed in the execution state - https://github.com/aws/aws-durable-execution-sdk-java/pull/508/changes#diff-61bd48c30220663a8125a529da3d92341c1e338f7cb04ededc7bdafe8fa0dd0aR36

ayushiahjolia: > nit: we're using isReplay in JS (I think it can still be confusing but replay is a terminology that is more commonly used around durable functions). I can be convinced otherwise

> Replay is applicable only to user functions in Context operations. I think we lack a term for the other operations.

I think isContinuation is more suitable here because it is just for tracking continuation of operation, it has nothing to do with replaying function.

examples/template.yaml
ayushiahjolia: The wrapper (`/opt/otel-handler`) attaches the ADOT auto-instrumentation agent, which creates its own TracerProvider. Our plugin also needs its own TracerProvider (because we need a custom deterministic IdGenerator for cross-invocation span correlation). Having two providers export to the same collector causes X-Ray to render them as separate/disconnected service nodes in the trace map.

In JS/Python, the plugin monkey-patches the agent's ID generator directly and shares the agent's provider - one provider, no conflict. In Java, the agent's provider is inaccessible so we can't swap its ID generator.

We still use the ADOT layer - it provides the OTLP collector extension that exports our spans to X-Ray. We just skip the wrapper/agent part since our plugin handles instrumentation directly.

sdk/src/main/java/software/amazon/lambda/durable/plugin/OperationInfo.java
SilanHe: >I think isContinuation is more suitable here because it is just for tracking continuation of operation, it has nothing to do with replaying function.

So I think we're comparing the clarity of isContinuation vs isReplay. I took a look at how each variable is used (JS vs Java) and it's identical.

isContinuation is true if the operation already exists in the ExecutionState. isContinuation is false for virtual child context and operations that have not started yet.

isReplay is true if the operation already exists in the ExecutionState. isReplay is false for virtual child context and operations that have not started yet.

So my understanding is that we're confusing the "internal" "Replay Mode" with this proposed `isReplay` variable that will be exposed to customers via the plugin interface. I think the meaning of the `isReplay` here is strictly referring to the onOperationStart hook. So whether we're replaying the execution of that hook. I think we all agree that we're not replaying the entire operation in this case.

In my opinion, `isContinuation` refers more to the operation as a whole which I do understand. isContinuation subtly implies that the onOperationStart hook will provide additional information or "continuation" on the subsequent call, which I don't believe is true. `isReplay` would technically be the more accurate term. Therefore, I still think `isReplay` is better. In addition, "Replay" is a terminology we already use elsewhere (e.g. https://docs.aws.amazon.com/durable-execution/getting-started/key-concepts/#replay).

## Context:

https://github.com/aws/aws-durable-execution-sdk-java/blob/main/sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java#L523

It's called as part of `fireOnOperationStart`. 

See where `fireOnOperationStart` is called: 

1. For virtual operations always `fireOnOperationStart(null)`: https://github.com/aws/aws-durable-execution-sdk-java/blob/main/sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java#L132

2. for operations which exist in ExecutionState call `fireOnOperationStart(existing)`: https://github.com/aws/aws-durable-execution-sdk-java/blob/main/sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java#L144

3. For operations which don't exist in ExecutionState call `fireOnOperationStart(null)`: https://github.com/aws/aws-durable-execution-sdk-java/blob/main/sdk/src/main/java/software/amazon/lambda/durable/operation/BaseDurableOperation.java#L163

SilanHe (changes requested):

sdk/src/main/java/software/amazon/lambda/durable/plugin/OperationInfo.java
```diff
@@ -27,4 +29,5 @@ public record OperationInfo(
         String subType,
         String parentId,
         Instant startTimestamp,
-        Instant endTimestamp) {}
+        Instant endTimestamp,
+        boolean isReplay) {}
```
zhongkechen, line 33: Why does this exist only in `OperationInfo`, not in `OperationEndInfo`?

sdk/src/main/java/software/amazon/lambda/durable/plugin/PluginInfoConverter.java
```diff
@@ -32,7 +32,8 @@ public static OperationInfo toOperationInfo(Operation operation, OperationIdenti
                 identifier.subType() != null ? identifier.subType().getValue() : null,
                 parentId,
                 operation != null ? operation.startTimestamp() : Instant.now(),
-                operation != null ? operation.endTimestamp() : null);
+                operation != null ? operation.endTimestamp() : null,
+                operation != null);
```
zhongkechen, line 36: Is `operation != null` true when the operation starts for the first time?

ayushiahjolia: it's the opposite. When the operation starts for the first time, operation is null, so isReplay = false. When the operation already exists in the execution state, operation is non-null, so isReplay = true.

sdk/src/main/java/software/amazon/lambda/durable/plugin/OperationInfo.java
ayushiahjolia: We only need isReplay in the start of the operation to decide span strategy, there is no use of isReplay in OperationEndInfo. 
Also, we are discussing about extending OperationInfo in OperationEndInfo - isReplay will then be available in OperationEndInfo as well in case we need it.

sdk/src/main/java/software/amazon/lambda/durable/plugin/PluginInfoConverter.java
zhongkechen: Ah, ok, it's fired before the operation is actually started.

sdk/src/main/java/software/amazon/lambda/durable/plugin/OperationInfo.java
zhongkechen: How does a plugin know if onOperationEnd is called for a replay of a context operation?

ayushiahjolia: For Otel plugin specififcally, we keep track of spans in operationSpans, but for general plugin, this might be an issue. Let me add it to OperationEndInfo as well

zhongkechen (approved):

SilanHe (approved):

Issue #463, opened 2026-06-17
Add FileSystem serdes


dhegberg:
Add FileSystem serdes for parity with JavaScript:
https://github.com/aws/aws-durable-execution-sdk-js/blob/main/packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts

zhongkechen:
Writing to S3 isn't synchronous and strongly consistent. The synchronization interval is 60s. Users using this feature may experience loss of data if Lambda runtime crashes.

Persisting payloads in SerDes isn't a valid use case of SerDes because the SerDes interface lacks enough information (e.g. operation id) to allow SerDes to identify the payload. We will need a different mechanism if we really want to support this use case.

zhongkechen:
Docs for FS SerDes in JS: https://docs.aws.amazon.com/durable-execution/sdk-reference/state/serialization/#filesystem-serdes
Design: https://github.com/aws/aws-durable-execution-sdk-java/blob/main/docs/adr/005-filesystem-serdes.md
- Option 1 is blocked by #509

PR #512, opened 2026-07-02
docs: add ADR for filesystem payload offloading

zhongkechen:
## Summary
- add ADR-005 for filesystem payload offloading design
- compare reusing SerDes with introducing a PayloadOffloader interface
- document extra package naming, caching, threading, exception handling, and AI recommendation

## Preview

https://github.com/aws/aws-durable-execution-sdk-java/blob/codex/adr-filesystem-payload-offloading/docs/adr/005-filesystem-serdes.md

## Issue
- Refs #463

## Testing
- not run; docs-only change

### DIFF
--- a/docs/adr/005-filesystem-serdes.md
+++ b/docs/adr/005-filesystem-serdes.md
@@ -1,0 +1,534 @@
+# ADR-005: Payload Offloading for Filesystem Storage
+
+**Status:** Proposed  
+**Date:** 2026-07-02
+
+## Context
+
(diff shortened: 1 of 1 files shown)


SilanHe (commented):
I like that we're discussing this decision in open source. 

So are you proposing `Approach B : Create a PayloadOffloader Interface` as is recommended by the AI?

zhongkechen:
> So are you proposing `Approach B : Create a PayloadOffloader Interface` as is recommended by the AI?

I think we go either A then B or directly go only B. Whichever we choose, B is the ultimate solution we want to have for payload offloading. We didn't implement B yet in the other SDKs due to time constrants. B will take more time to implement than A in other SDKs. However the cost difference between these two is much smaller in Java SDK

ayushiahjolia (approved):

PR #552, opened 2026-07-20
feat(conformance): add Java conformance test package (step, wait suites)

wangyb-A:
## Summary

Adds `conformance-tests/` (artifact `aws-durable-execution-sdk-java-conformance-tests`): Java handlers + SAM templates for the **step** and **wait** conformance suites, exercised by the language-agnostic runner [`aws-durable-execution-conformance-tests`](https://github.com/aws/aws-durable-execution-conformance-tests) (pinned `0.1.0`). Java counterpart of aws/aws-durable-execution-sdk-js#733.

- 20 step handlers + 5 wait handlers, using the SDK's real API — retry counting uses the built-in `StepContext.getAttempt()` (no DynamoDB side channel)
- `template_step.yaml` / `template_wait.yaml` map each handler to requirement IDs via `TestingMetadata.TestDescription`
- `scripts/inject_execution_role.py` (ported from the JS package): CI points every function at the pre-existing `TEST_LAMBDA_EXECUTION_ROLE_ARN` and drops the self-created `DurableFunctionRole`; the checked-in templates stay self-contained for local runs
- `.github/workflows/conformance-tests.yml`: one parallel matrix job per suite (`matrix.suite: [step, wait]`), OIDC via the existing `TEST_ROLE_ARN` / `TEST_ACCOUNT_ID` secrets (same as e2e-tests). Add future suites by extending the matrix and shipping `template_<suite>.yaml` + handlers.

## Testing

- `mvn -B -Dmaven.test.skip=true install` at repo root: green (module included, spotless-formatted)
- Full local runner execution against a live account, both suites:
  - **step: 20/20 PASSED**
  - **wait: 5/5 PASSED**
- `inject_execution_role.py` unit-smoke-tested: rewrites all functions, removes `DurableFunctionRole`

### DIFF
--- a/.github/workflows/conformance-tests.yml
+++ b/.github/workflows/conformance-tests.yml
@@ -10,6 +10,8 @@ on:
   pull_request:
     branches: [main]
     paths:
+      - "sdk/**"
+      - "pom.xml"
       - "conformance-tests/**"
       - ".github/workflows/conformance-tests.yml"
   workflow_dispatch:
--- a/.github/workflows/conformance-tests.yml
+++ b/.github/workflows/conformance-tests.yml
@@ -1,0 +1,111 @@
+name: Conformance Tests
+
+# Full-integration conformance run: builds the SDK + Java handlers, installs the
+# pinned language-agnostic runner from PyPI, then deploys + invokes + validates
+# one SAM stack per suite. Each suite runs as its own parallel matrix job. To add
+# a suite later, add its name to `strategy.matrix.suite` (and ship its
+# template_<suite>.yaml + handlers under conformance-tests/).
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - "conformance-tests/**"
+      - ".github/workflows/conformance-tests.yml"
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.head_ref || github.ref_name || github.run_id }}-conformance
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  id-token: write # Required for AWS OIDC credentials
+
+jobs:
+  conformance:
+    name: conformance (${{ matrix.suite }})
+    runs-on: ubuntu-latest
+    env:
+      AWS_REGION: us-west-2
+    strategy:
+      fail-fast: false
+      matrix:
+        suite: [step, wait]
+    defaults:
+      run:
+        working-directory: conformance-tests
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v7
+
+      - name: Setup Java 21
+        uses: actions/setup-java@v5
+        with:
+          distribution: corretto
+          java-version: 21
+          cache: maven
+
+      - name: Setup AWS SAM CLI
+        uses: aws-actions/setup-sam@v3
+        with:
+          use-installer: true
+
+      - name: Build SDK and conformance handlers
+        working-directory: .
+        run: mvn -B -q -Dmaven.test.skip=true install --file pom.xml
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.14"
+
+      - name: Install conformance runner
+        run: pip install aws-durable-execution-conformance-tests==0.1.0
+
+      - name: Configure AWS credentials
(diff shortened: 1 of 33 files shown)


.github/workflows/conformance-tests.yml
```diff
@@ -0,0 +1,111 @@
...
+          # Point every function at the pre-existing execution role and drop the
+          # self-created DurableFunctionRole. Mutates only the CI checkout; the
+          # checked-in template stays self-contained for local runs.
+          python3 scripts/inject_execution_role.py \
+            --template template_${{ matrix.suite }}.yaml \
+            --role-arn "$ROLE_ARN"
+
+      - name: Run conformance suite
+        run: |
+          python -m aws_durable_execution_conformance_tests.app \
```
zhongkechen, line 95: is `validate` command preferred? https://github.com/aws/aws-durable-execution-conformance-tests/blob/main/pyproject.toml#L51

conformance-tests/template_step.yaml
```diff
@@ -0,0 +1,374 @@
+AWSTemplateFormatVersion: '2010-09-09'
```
zhongkechen, line 1: Can we use a script to generate sam templates from code?

.github/workflows/conformance-tests.yml
```diff
@@ -0,0 +1,111 @@
...
+# pinned language-agnostic runner from PyPI, then deploys + invokes + validates
+# one SAM stack per suite. Each suite runs as its own parallel matrix job. To add
+# a suite later, add its name to `strategy.matrix.suite` (and ship its
+# template_<suite>.yaml + handlers under conformance-tests/).
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - "conformance-tests/**"
```
zhongkechen, line 15: sdk changes should trigger the conformance tests.

conformance-tests/template_step.yaml
wangyb-A: We need to include the testing metadata in the template so the mapping from handler -> test requirements has to be somewhere.

```
    TestingMetadata:
      TestDescription: ["1-1"]
```

.github/workflows/conformance-tests.yml
wangyb-A: `validate` there is a hatch env-script alias (`[tool.hatch.envs.default.scripts]`), defined as exactly `python -m aws_durable_execution_conformance_tests.app {args}` — it only exists when running `hatch run` inside a checkout of the runner repo. The published wheel (`aws-durable-execution-conformance-tests==0.1.0`) declares no `[project.scripts]` console entry point, so after `pip install` there is no `validate` command on PATH. CI installs from PyPI, hence the `python -m` form (same as the JS conformance workflow). If the runner repo later adds a console entry point, we can switch.

conformance-tests/template_step.yaml
wangyb-A: Or we can find a way to include that in the handler / separate file, e.g.

`@ConformanceTest("1-1")` or a small `manifest.yaml`

.github/workflows/conformance-tests.yml
wangyb-A: fixed in next rev

.github/workflows/conformance-tests.yml
wangyb-A: Created an issue for this: https://github.com/aws/aws-durable-execution-conformance-tests/issues/11

zhongkechen: I added one in my PR https://github.com/aws/aws-durable-execution-conformance-tests/blob/c5fd3faad26685cdb185c100da544a79916e2f83/packages/aws-durable-execution-conformance-tests/pyproject.toml#L28

conformance-tests/template_step.yaml
zhongkechen: or a naming convention could help: `1-1 StepBasic.java`

zhongkechen: We can do this later. https://github.com/aws/aws-durable-execution-sdk-java/issues/553

zhongkechen (approved):

PR #555, opened 2026-07-22
feat(conformance): add map suite

wangyb-A:
## Summary

Adds the **map** conformance suite (13 handlers), completing suite coverage. Follow-up to #552/#554.

- **`MapEmpty` (9-4) is not included** — its test case lands in a separate CR; until then 9-4 reports as non-blocking `UNCOVERED`.
- The 6 pre-existing documentation-comment gaps (9-2, 9-6, 9-13, 9-14, 9-19, 9-20 — Java SDK API limitations) are formalized as `NotImplemented` declarations.
- Declarations live on `MapBasic`'s `TestingMetadata` (the runner supports top-level and per-resource forms, but CloudFormation rejects unknown **top-level** sections — learned via a failed changeset).
- CI matrix now runs all 9 suites, and the workflow reuses a stable per-suite stack with `--no-cleanup` (runs serialized globally to avoid concurrent updates on shared stacks).

## Testing

Live validation of this template: **13 PASSED, 0 FAILED** (remaining map requirements report as NOT_IMPLEMENTED/UNCOVERED, non-blocking).


nvasiu (approved):

conformance-tests/src/main/java/map/MapToleratedExceeded.java
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * 9-9: Map tolerated-failure-count exceeded (stops early). Items 0 and 1 fail (failure count 2 exceeds the tolerance of
+ * 1), so item 2 is never started.
+ */
+public class MapToleratedExceeded extends DurableHandler<Object, Map<String, Object>> {
+
+    @Override
+    public Map<String, Object> handleRequest(Object input, DurableContext context) {
+        var config = MapConfig.builder()
+                .maxConcurrency(1)
```
zhongkechen, line 23: Why did we use maxConcurrency=1 in most of our cases

Issue #529, opened 2026-07-10
[Bug]: Missing checkpoint when map operation has empty items


wangyb-A:
### Expected Behavior

When a map receives empty items, I expected a operation started and succeeded checkpoint.

### Actual Behavior

When a map receives empty items, no checkpoint is done.

### Steps to Reproduce

```

package map;

import java.util.List;
import software.amazon.lambda.durable.DurableContext;
import software.amazon.lambda.durable.DurableHandler;
import software.amazon.lambda.durable.config.MapConfig;
import software.amazon.lambda.durable.model.MapResult;

public class MapEmpty extends DurableHandler<Object, List<String>> {

    @Override
    public List<String> handleRequest(Object input, DurableContext context) {
        MapResult<String> result = context.map(
                "empty", List.<String>of(), String.class, (item, index, ctx) -> item, MapConfig.builder().build());
        return result.results();
    }
}


```

### SDK Version

2.0.0

### Java Version

17

### Is this a regression?

No

zhongkechen:
The real bug is the impact on replay status and plugin behavior:
- context.isReplaying() to become false prematurely, so replay logs or guarded side effects may run again
- Plugin lifecycle imbalance: onOperationStart fires, but onOperationEnd does not; OTel may report the completed map as PENDING.

nvasiu:
Summary of the above PR:
- Added a fix for empty maps not checkpointing, but introduces breaking changes, so the fix is behind a DurableConfig flag that users have to opt in to.
- The default behaviour (empty maps do not checkpoint) remains the same.
- In a future major version we will update the fixed behaviour to be default, and remove the flag.

PR #556, opened 2026-07-22
test(conformance): add map 9-4 (empty items list)

wangyb-A:
## Summary

Adds the single remaining map test case: **`MapEmpty` (9-4)** — map invoked with an empty items list. Follow-up to #555, which shipped the map suite without it.

## ⚠️ This test is expected to FAIL

The current SDK diverges from the requirement: an empty-items map completes with `InvocationCompleted` **without** emitting `ContextStarted`/`ContextSucceeded`. Per the conformance suite's follow-the-requirement principle, the handler uses the real `context.map` API and the red test documents the SDK bug — it should stay red until either the SDK behavior is fixed or the requirement is amended.

Expect the **map** CI job on this PR to fail on exactly this requirement.

## Testing

Live validation of this template: **13 PASSED, 1 FAILED (9-4, the expected divergence), 6 NOT_IMPLEMENTED**. Failure detail:

```
9-4: Event[EventId=2].EventType: expected 'ContextStarted', got 'InvocationCompleted'
```

### DIFF
--- a/conformance-tests/src/main/java/map/MapEmpty.java
+++ b/conformance-tests/src/main/java/map/MapEmpty.java
@@ -1,0 +1,24 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package map;
+
+import java.util.List;
+import software.amazon.lambda.durable.DurableContext;
+import software.amazon.lambda.durable.DurableHandler;
+import software.amazon.lambda.durable.config.MapConfig;
+import software.amazon.lambda.durable.model.MapResult;
+
+/** 9-4: Map with an empty items list completes immediately with an empty results list. */
+public class MapEmpty extends DurableHandler<Object, List<String>> {
+
+    @Override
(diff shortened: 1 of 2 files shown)


wangyb-A:
blocked by #529

SilanHe (approved):

conformance-tests/src/main/java/map/MapEmpty.java
```diff
@@ -0,0 +1,24 @@
...
+package map;
+
+import java.util.List;
+import software.amazon.lambda.durable.DurableContext;
+import software.amazon.lambda.durable.DurableHandler;
+import software.amazon.lambda.durable.config.MapConfig;
+import software.amazon.lambda.durable.model.MapResult;
+
+/** 9-4: Map with an empty items list completes immediately with an empty results list. */
+public class MapEmpty extends DurableHandler<Object, List<String>> {
```
zhongkechen, line 12: A new config for empty map must be enabled

zhongkechen (changes requested):

wangyb-A:
Done in #560

PR #559, opened 2026-07-23
test: add plugin conformance suite handlers (10-1..10-7)

wangyb-A:
## What

7 handlers + shared `ConformanceLoggingPlugin`/`FaultyConformancePlugin` for the `plugin` conformance suite in `conformance-tests/src/main/java/plugin/`, plus `template_plugin.yaml`.

## Validatio

CI is passing

### DIFF
--- a/conformance-tests/src/main/java/plugin/ConformanceLoggingPlugin.java
+++ b/conformance-tests/src/main/java/plugin/ConformanceLoggingPlugin.java
@@ -44,21 +44,21 @@ public class ConformanceLoggingPlugin implements DurableExecutionPlugin {
     @Override
     public void onOperationStart(OperationInfo info) {
         if (isStep(info.type())) {
-            System.out.println(prefix + " operation-start");
+            System.out.println(prefix + " operation-start op=" + info.id());
         }
     }
 
     @Override
     public void onOperationEnd(OperationEndInfo info) {
         if (isStep(info.type())) {
-            System.out.println(prefix + " operation-end status=" + info.status());
+            System.out.println(prefix + " operation-end op=" + info.id() + " status=" + info.status());
         }
     }
 
     @Override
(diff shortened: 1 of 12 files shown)


zhongkechen:
We found a disparity in the plugin interface: https://github.com/aws/aws-durable-execution-sdk-python/pull/591/changes#r3661794243 

Are we able to detect this in this test suite?

conformance-tests/src/main/java/plugin/ConformanceLoggingPlugin.java
```diff
@@ -0,0 +1,93 @@
...
+
+    /** Returns {@code , "durableExecutionArn": "<arn>"} when captured, otherwise an empty string. */
+    private String arnField() {
+        return executionArn == null ? "" : String.format(", \"durableExecutionArn\": \"%s\"", executionArn);
+    }
+
+    @Override
+    public void onInvocationStart(InvocationInfo info) {
+        this.executionArn = info.durableExecutionArn();
+        System.out.println(String.format(
```
zhongkechen, line 45: We can use MDC to inject logging metadata

wangyb-A:
> Are we able to detect this in this test suite?

We should, there wil be more tests coming, tracked in https://github.com/aws/aws-durable-execution-conformance-tests/issues/5

conformance-tests/src/main/java/plugin/ConformanceLoggingPlugin.java
wangyb-A: Yes that's feasible. However I chose to use raw print here to make sure the message field and durableExecutionArn is at top level without extra log configuration, and these tests only aims to test the plugin interface. We can use MDC in the tests for logging IMO

ayushiahjolia (approved):

PR #560, opened 2026-07-23
Fix missing checkpoint from map operation with empty items

nvasiu:

### Issue Link, if available

https://github.com/aws/aws-durable-execution-sdk-java/issues/529

### Description

Currently when a map operation gets an empty list of items, it will complete locally without producing the usual durable checkpoints. This resulted in logs or side effects re-running unintentionally, or plugin hooks not being called.

This PR handles empty maps as first-class checkpointed operations, so behaviour is consistent with non-empty maps.

### Code Changes

`MapOperation.java`
- `start()` no longer calls `markAlreadyCompleted()` on empty maps. 
  - It now sends a start checkpoint to the backend, like in the non-empty case. 
    - The sync `sendOperationUpdate()` is used to guarantee the start checkpoint is acknowledged before sending the success checkpoint.
  - And then it calls `handleCompletion()`, to mimic how non-empty maps send a success checkpoint to the backend.

- `replay()` no longer guards against empty maps.
  - The guard existed because previously empty maps did not checkpoint, causing this method to hang. But now that empty maps checkpoint, `replay()` will be able to flow through the normal SUCCEEDED path.

- `get()` no longer returns an empty result for empty maps.
  - So now the actual checkpointed result will be returned via `join()`, consistent with the normal path.

`MapIntegrationTest.java`
- Updated the empty map tests to now expect the correct number of events (previously they expected 0, because all events were skipped for empty maps).
- Added a new test to check that empty maps are correctly replayed.

`PluginIntegrationTest.java`
- Added a test to check that an empty map operation will properly fire the `onOperationStart` and `onOperationEnd` hooks.

---

**Also contains a commit for the relevant conformance suite test case. See this PR: https://github.com/aws/aws-durable-execution-sdk-java/pull/556.**

---

### Checklist

#### Unit Tests

Have unit tests been written for these changes? No unit tests for map operation.

#### Integration Tests

Have integration tests been written for these changes? YES

#### Examples

### DIFF
--- a/conformance-tests/src/main/java/map/MapEmpty.java
+++ b/conformance-tests/src/main/java/map/MapEmpty.java
@@ -1,0 +1,24 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+package map;
+
+import java.util.List;
+import software.amazon.lambda.durable.DurableContext;
+import software.amazon.lambda.durable.DurableHandler;
+import software.amazon.lambda.durable.config.MapConfig;
+import software.amazon.lambda.durable.model.MapResult;
+
+/** 9-4: Map with an empty items list completes immediately with an empty results list. */
+public class MapEmpty extends DurableHandler<Object, List<String>> {
+
+    @Override
+    public List<String> handleRequest(Object input, DurableContext context) {
+        MapResult<String> result = context.map(
+                "empty",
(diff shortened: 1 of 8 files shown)


SilanHe (approved):

nvasiu:
New commit to put the new fixed behaviour behind a DurableConfig flag, so we don't introduce a breaking change.

We will make the fixed behaviour default, and remove the flag in a future major version release.

PR #561, opened 2026-07-24
ci: use latest conformance suite (git main) in conformance CI

wangyb-A:
Switches the conformance-tests workflow to install the runner from git main (subdirectory packages/aws-durable-execution-conformance-tests) instead of the pinned ==0.1.0 PyPI release, so new upstream suites/requirements are exercised automatically. Draft until reviewed.


conformance-tests/src/main/java/child/ChildLargePayload.java
```diff
@@ -3,17 +3,25 @@
...
 import software.amazon.lambda.durable.DurableHandler;
+import software.amazon.lambda.durable.logging.LoggerConfig;
 
 /** 3-11: Child context large payload (ReplayChildren mode) */
 public class ChildLargePayload extends DurableHandler<Object, String> {
 
+    @Override
+    protected DurableConfig createConfiguration() {
+        return DurableConfig.builder()
+                .withLoggerConfig(LoggerConfig.withReplayLogging())
```
zhongkechen, line 17: disabled replay-aware logging so that `child.getLogger().info` can always write logs without checking replaying status

zhongkechen (approved):

PR #574, opened 2026-07-29
Add plugin-suite conformance handlers for requirements 10-8..10-18

wangyb-A:
## Summary
Adds 9 plugin-suite conformance handlers (`conformance-tests/src/main/java/plugin/PluginOperationChange.java` through `PluginSuspensionInvocationEnd.java` / `PluginFaultyAndHealthy.java`, plus shared `PluginSupport`) implementing edge-case requirements 10-8..10-13 and 10-15..10-17 from aws/aws-durable-execution-conformance-tests#43 (land that PR first — the pinned runner must carry the requirement YAMLs before CI can resolve these IDs).

Rebased on `main`, which independently landed the 10-1..10-7 plugin backfill and the suite-discovery CI matrix: this branch only extends `template_plugin.yaml` with the new function entries and carries no CI workflow changes — the `plugin` suite is picked up by discovery.

All handlers use real SDK APIs only: real `WaitStrategies`/`StepConfig` retries, real waits, real `context.parallel` / child contexts, real `DurableConfig.builder().withPlugins(...)` registration — no mocked behavior. Records are single-line JSON stdout stamped with `durableExecutionArn`. 10-16 stamps the same-invocation `first` flag (from `InvocationEndInfo.isFirstInvocation()`), closing a reviewer-reported match-ambiguity hole.

## Deliberately excluded: 10-14 and 10-18 (known Java SDK gaps)
Handlers for these two requirements are NOT included in this PR because the Java plugin API cannot satisfy them; both were reproduced with live cross-SDK evidence (Python and JS pass both):

| Gap | Requirement | Detail |
|---|---|---|
| 1 | 10-14 | `OperationEndInfo` exposes no serialized-result accessor — plugins cannot see an operation's checkpointed result at operation-end (error side works via `error()`). |
| 2 | 10-18 | `onOperationStart` with `isReplay=true` fires only for STEP/CONTEXT on non-terminal replay — not for WAIT (nor INVOKE/CALLBACK). A wait still pending during a replayed invocation is never reported to plugins. |

Until the SDK closes these gaps, full-suite runs will report the two IDs as uncovered for Java; handlers can be added back alongside the SDK fixes.

## Validation (us-west-2, live deploy)
All 9 included requirements PASS (10-8..10-13, 10-15..10-17); the 10-1..10-7 backfill from main is unchanged.


wangyb-A:
The excluded test cases will be published in separate CR

conformance-tests/src/main/java/plugin/PluginSupport.java
```diff
@@ -0,0 +1,50 @@
...
+    private PluginSupport() {}
+
+    /** Operation type token for step operations as reported by {@code OperationInfo#type()} (AWS SDK {@code OperationType}). */
+    static boolean isStep(String type) {
+        return "STEP".equals(type);
+    }
+
+    /** Operation type token for wait operations. */
+    static boolean isWait(String type) {
+        return "WAIT".equals(type);
```
nvasiu, line 24: Nit: `isWait()` and `isStep()` are case sensitive so you need to do `.toUpperCase()` when passing to them. But `isStepChange()` ignores the case, so that's not necessary.

nvasiu (approved):

Issue #53, opened 2026-02-05
[Bug]: checkpoint call doesn't fetch paginated operations


zhongkechen:
### Expected Behavior

CheckpointDurableExecution API will return a list of operations with a nextMarker. If nextMarker exists, the SDK should fetch next pages of operations

### Actual Behavior

The next pages of operations are not returned and processed

https://github.com/aws/aws-durable-execution-sdk-java/blob/main/sdk/src/main/java/com/amazonaws/lambda/durable/execution/CheckpointBatcher.java#L91

### Steps to Reproduce

N/A

### SDK Version

all

### Java Version

17

### Is this a regression?

No

PR #583, opened 2026-07-31
fix(otel): set explicit span status OK on successful operation and attempt spans

SilanHe:
## Summary

set explicit span status OK on successful operation and attempt spans

## Testing

- otel-plugin module build — clean
- otel-plugin tests — 168/168 (0 failures/errors), incl. 4 new OK-on-success tests (InvocationOtelPluginTest 37→39, ExecutionOtelPluginTest 23→25)

## Coordination

Aligns Java with the operation-status/span-status model tracked by aws/aws-durable-execution-conformance-tests#52 and #53, matching Python #604 and the JS companion change.


SilanHe:
Addressed the Codex P1 / Claude finding in `ad07ddb`.

The `else -> setStatus(StatusCode.OK)` on the operation and continuation spans was gating on `info.error() == null`, which mislabeled non-success terminal states as OTel `OK`: `onOperationEnd` fires for every terminal status, and `extractErrorFromOperation` returns null for `CANCELLED` (always) and for `FAILED`/`TIMED_OUT`/`STOPPED` with no error object. All four operation/continuation sites in `InvocationOtelPlugin.java` and `ExecutionOtelPlugin.java` now gate on `"SUCCEEDED".equals(info.status()) || info.status() == null`. The `|| info.status() == null` clause preserves `OK` for successful statusless virtual (FLAT CONTEXT) operations — verified against `ChildContextOperation` (success fires `fireOnOperationEnd(null, null, false)`; failure passes a non-null error and hits the ERROR branch). Non-success statuses stay `UNSET`. Attempt-span `else if (info.succeeded())` left untouched.

Added 6 regression tests (3 per plugin): non-`SUCCEEDED` terminal status with null error -> NOT `OK` (same-invocation + continuation paths), plus a null-status virtual-context -> `OK` case. otel-plugin module: 142/142 passing.

zhongkechen (approved):

PR #585, opened 2026-07-31
feat: add custom map iteration naming

### LINKED ISSUE #528: [Feature]: Item namer for map operation
### What would you like?

I would like to have a item namer for map operation

### Possible Implementation

python implementation for example: https://github.com/aws/aws-durable-execution-sdk-python/pull/387

No

### Does this require an RFC?

No

### Additional Context

_No response_

wangyb-A:
## Summary
- add `MapConfig.itemNamer` for custom nested map-iteration names
- add a typed `itemNamer(Class<I>, BiFunction<? super I, Integer, String>)` overload so domain-object namers need no cast
- preserve a `null` namer result as an unnamed iteration
- resolve and validate names before allocating the map operation ID
- reject `itemNamer` with `NestingType.FLAT`, whose virtual iterations have no context operation to name
- detect changed iteration names when replaying STARTED and cached SUCCEEDED maps
- implement Java conformance requirement 9-13

## Item namer API
Untyped form (unchanged):

```java
MapConfig.builder()
        .itemNamer((item, index) -> item + "-" + index)
        .build();
```

Typed form for domain objects:

```java
MapConfig.builder()
        .itemNamer(Order.class, (order, index) -> order.id())
        .build();
```

The typed overload bridges through `Class#cast`, so it needs no `@SuppressWarnings("unchecked")` and a mismatched item class fails with a `ClassCastException` naming the offending type. `MapConfig` stays non-generic, so no raw-type warnings are introduced for existing callers.

## Compatibility
- `MapConfig` remains non-generic
- the original public `MapOperation` constructor is retained as a delegating overload
- existing `DurableContext` method signatures are unchanged
- default map iteration naming is unchanged when no item namer is configured

## Validation
- `mvn -pl sdk test` — 1110 passed
- `mvn -pl sdk-integration-tests test` — 393 passed
- `mvn -pl conformance-tests package -DskipTests` — passed
- `mvn spotless:check` — passed
- `javac -Xlint:unchecked,rawtypes` — no warnings in the item-namer files
- real deployed map conformance run is pending

Closes #528


sdk/src/main/java/software/amazon/lambda/durable/config/MapConfig.java
```diff
@@ -105,6 +125,45 @@ public Builder nestingType(NestingType nestingType) {
...
+         *
+         * <p>Each item is passed through {@link Class#cast}, so supplying a class that does not match the map's items
+         * fails with a {@link ClassCastException} naming the offending type.
+         *
+         * @param itemType the class of the map's items
+         * @param itemNamer the item namer, or null to use default iteration naming
+         * @param <I> the map item type accepted by the namer
+         * @return this builder for method chaining
+         */
+        public <I> Builder itemNamer(Class<I> itemType, BiFunction<? super I, Integer, String> itemNamer) {
```
zhongkechen, line 161: Would an overload with `TypeToken<I>` be necessary?

wangyb-A: Don't think it's necessary here, because a TypeToken overload adds no runtime safety here (no cast available). Do you have any example that TypeToken will be helpful ?

zhongkechen (approved):

PR #594, opened 2026-08-05
feat(otel): OtelPluginConfig parameter to replace constructor overloads for configuration options

SilanHe:
https://github.com/aws/aws-durable-execution-conformance-tests/issues/23#issuecomment-5184325293

Replaces the OTel plugins' telescoping constructors with an immutable `OtelPluginConfig` builder and a config-only constructor, and uses a single `ProviderSource` enum (`EXPLICIT`/`GLOBAL`/`AUTO_OTLP`, default `GLOBAL`) to select the tracer provider; pass `AUTO_OTLP` for a plugin-owned OTLP/HTTP provider. Breaking (preview API). All tests pass.

### DIFF
--- a/otel-plugin/src/main/java/software/amazon/lambda/durable/otel/ExecutionOtelPlugin.java
+++ b/otel-plugin/src/main/java/software/amazon/lambda/durable/otel/ExecutionOtelPlugin.java
@@ -162,18 +162,24 @@ public class ExecutionOtelPlugin implements DurableExecutionPlugin {
     /**
      * Creates a Workflow-rooted OTel plugin from configuration alone (no caller-supplied tracer provider builder).
      *
-     * <p>The provider is resolved from {@link OtelPluginConfig#resolveSource()}: {@link ProviderSource#GLOBAL} when
-     * {@code useDefaultTracerProvider(true)} (the ADOT/global provider), otherwise the default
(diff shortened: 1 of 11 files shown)


otel-plugin/src/main/java/software/amazon/lambda/durable/otel/ExecutionOtelPlugin.java
```diff
@@ -150,14 +150,33 @@ public ExecutionOtelPlugin(SdkTracerProviderBuilder tracerProviderBuilder, Conte
...
-     * @param enableMdc if true, injects traceId/spanId/otelTraceSampled into SLF4J MDC for log correlation
+     * @param enableMdc if true, injects traceId/spanId/traceSampled into SLF4J MDC for log correlation
      * @param workflowSpanName the name for the Workflow root span
      */
     public ExecutionOtelPlugin(
             SdkTracerProviderBuilder tracerProviderBuilder,
             ContextExtractor contextExtractor,
             boolean enableMdc,
             String workflowSpanName) {
+        this(tracerProviderBuilder, contextExtractor, enableMdc, workflowSpanName, INSTRUMENTATION_NAME);
```
zhongkechen: Do we need builder pattern here?

SilanHe: yeah I ended up adding a "config" object that gets instantiated through the builder pattern

zhongkechen (approved):

PR #606, opened 2026-08-07
feat: load plugins dynamically from Lambda layers

### LINKED ISSUE #605: [Feature]: Load plugins dynamically from Lambda layers
### What would you like?

I would like the SDK to support opt-in discovery and registration of
`DurableExecutionPlugin` implementations from runtime dependencies during cold
start.

The target Lambda workflow is:

1. Publish a plugin JAR in a Lambda layer under `java/lib`.
2. Attach the layer to an existing function.
3. Set configuration such as
   `DURABLE_EXECUTION_PLUGINS=otel,com.example.audit`.
4. Have the SDK load and register those plugins without rebuilding or
   repackaging the function's application artifact.

Today plugins must be instantiated in application code and passed to
`DurableConfig.builder().withPlugins(...)`. Making a JAR available on the
Lambda class path is not sufficient, and Java-agent extension discovery is
separate from Durable Execution plugin registration.

The dynamic path should:

- Be explicitly enabled; an unset configuration must preserve current behavior.
- Load plugins once during cold-start configuration.
- Preserve configured ordering and support multiple plugins.
- Resolve providers from the normal application and Lambda-layer class path.
- Validate provider type and SDK/API compatibility with actionable failures.
- Define duplicate handling when a plugin is both explicitly and dynamically
  registered.
- Preserve `withPlugins(...)` for applications that prefer code-based
  registration.

### Possible Implementation

Add a provider/factory SPI, for example `DurableExecutionPluginProvider`, and
discover implementations with `ServiceLoader`. Providers could expose a stable
name and create a `DurableExecutionPlugin`, while the environment variable acts
as an explicit allow-list rather than automatically executing every provider
found on the class path.

An alternative is supporting fully qualified class names with reflection, but a
provider interface would avoid requiring every plugin to have a public no-arg
constructor and would give the SDK a place to perform compatibility validation.

Provider-specific settings could remain in namespaced environment variables.

No

### Does this require an RFC?

Yes

### Additional Context

The immediate use case is distributing observability integrations, including
OpenTelemetry, independently from the function artifact. The same capability is
useful for logging, metrics, auditing, and other instrumentation plugins.

Attaching a layer and updating environment variables is still a Lambda
configuration deployment, but it avoids rebuilding and republishing application
code solely to add or update instrumentation.

zhongkechen:

### Issue Link, if available

Closes #605

### Description

- Add a versioned `DurableExecutionPluginProvider` SPI discovered with `ServiceLoader` when `DURABLE_EXECUTION_PLUGINS` is set.
- Load environment-selected plugins first in configured order, followed by plugins registered through `DurableConfig.withPlugins(...)`; both sources remain additive.
- Add `otel-invocation` and `otel-execution` providers for `InvocationOtelPlugin` and `ExecutionOtelPlugin`.
- Validate provider names, duplicates, API compatibility, plugin types, and creation failures with actionable configuration errors.
- Document Lambda layer packaging, provider authoring, selector names, and OTel layer configuration.

### Demo/Screenshots

Not applicable; this is SDK configuration and plugin discovery behavior.

### Checklist

#### Unit Tests

Yes. Added loader coverage for ordering, additive duplicate types, allow-list behavior, malformed configuration, discovery failures, compatibility validation, and plugin creation failures. Added provider tests for both OTel plugins.

#### Integration Tests

Yes. Added lifecycle coverage confirming dynamically loaded and explicitly configured plugins both receive events.

Also ran `mvn clean install` successfully across all reactor modules.

#### Examples

No new example was required. The configuration guide and OTel plugin README now include Lambda layer and environment-variable examples.

### DIFF
--- a/sdk/src/main/java/software/amazon/lambda/durable/DurableConfig.java
+++ b/sdk/src/main/java/software/amazon/lambda/durable/DurableConfig.java
@@ -103,6 +103,7 @@ public final class DurableConfig {
     private final PluginRunner pluginRunner;
 
     private DurableConfig(Builder builder) {
+        var plugins = DynamicPluginLoader.loadConfiguredPlugins(builder.plugins);
         this.durableExecutionClient = Objects.requireNonNullElseGet(
                 builder.durableExecutionClient, DurableConfig::createDefaultDurableExecutionClient);
         this.serDes = Objects.requireNonNullElseGet(builder.serDes, JacksonSerDes::new);
@@ -113,7 +114,7 @@ public final class DurableConfig {
         this.checkpointDelay = Objects.requireNonNullElseGet(builder.checkpointDelay, () -> Duration.ofSeconds(0));
         this.deserializeAfterSerialization = builder.deserializeAfterSerialization;
         this.checkpointEmptyMap = builder.checkpointEmptyMap;
-        this.pluginRunner = builder.plugins.isEmpty() ? PluginRunner.noOp() : new PluginRunner(builder.plugins);
+        this.pluginRunner = plugins.isEmpty() ? PluginRunner.noOp() : new PluginRunner(plugins);
 
         validateConfiguration();
     }
@@ -216,7 +217,7 @@ public final class DurableConfig {
     /**
      * Gets the plugin runner that dispatches lifecycle events to registered plugins.
      *
-     * <p>Returns a no-op runner if no plugins were registered via the builder.
+     * <p>Returns a no-op runner if no plugins were registered via the builder or loaded dynamically.
      *
      * @return PluginRunner instance (never null)
      * @deprecated This is a preview API that is experimental and may be changed or removed in future releases.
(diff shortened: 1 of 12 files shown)


sdk/src/main/java/software/amazon/lambda/durable/DurableConfig.java
```diff
@@ -216,7 +217,7 @@ public boolean shouldCheckpointEmptyMap() {
     /**
      * Gets the plugin runner that dispatches lifecycle events to registered plugins.
      *
-     * <p>Returns a no-op runner if no plugins were registered via the builder.
+     * <p>Returns a no-op runner if no plugins were registered via the builder or loaded dynamically.
```
SilanHe, line 220: nit: This comment is wrong. If we're returning a noOp runner then it means that no plugins were loaded dynamically

SilanHe (approved):

sdk/src/main/java/software/amazon/lambda/durable/DurableConfig.java
zhongkechen: if env var is empty and plugins are specified in DurableConfig.Builder, a real runner will be created.

Issue #604, opened 2026-08-07
Plugin hook info records missing state/metadata fields at invocation, attempt, and change hooks


wangyb-A:
## Summary
Java's plugin info records lack metadata/state fields the Python/JS SDKs expose. Live cross-SDK conformance evidence exists per gap.

**GA scope update:** payload surfaces (execution input/result, operation `result`) have been moved OUT of GA conformance scope — experimental, no longer asserted by the field-shape requirements (see aws/aws-durable-execution-conformance-tests#72). They remain listed at the end as post-GA parity items; the operation-result gap is still evidenced by the untouched behavioral requirement 10-14 (#581).

## In-scope missing fields (verified against `sdk/src/main/java/software/amazon/lambda/durable/plugin/`)
**`InvocationInfo`** — missing `operations` and `updatedOperations` maps (JS exposes both on every invocation hook).
**`InvocationEndInfo`** — missing the `operations` map AND `executionStartTime` (present on the start info but dropped from the end record — confirmed by live probe).
**`UserFunctionStartInfo`/`UserFunctionEndInfo`** (attempt hooks) — missing the operation-level `isReplay` indicator (Python and JS carry it on attempt infos; Java only has the distinct `isReplayingChildren`).
**`OperationChangeItemInfo`** — missing `attempt` and a replay indicator (JS/Python change items are full operation infos).
Secondary: `UserFunctionEndInfo` uses `boolean succeeded` where JS/Python use a SUCCEEDED/FAILED outcome enum.

## Evidence
Field-shape requirements 10-19..10-23 (aws/aws-durable-execution-conformance-tests#72), canonical-dump handlers in #602 — live run 18/21: 10-19 (maps + end-side start time), 10-21 (`isReplay` presence), 10-22 (item attempt/replay) red; 10-20 and 10-23 green. Notably green: end-info `isFirstInvocation` (ahead of JS), `OperationInfo.status` (#597), attempt-end timestamps (ahead of JS), `isReplayingChildren` semantics.

## Post-GA / experimental parity items (not currently asserted)
- `executionInput`/`executionResult` on invocation infos; `result` on `OperationInfo`/`OperationEndInfo`/`OperationChangeItemInfo` (the OperationEndInfo result gap remains tracked by behavioral test 10-14 / #581)

## Suggested fix
[...]

PR #618, opened 2026-08-10
fix(plugin): surface missing state fields on hook info records

wangyb-A:
## Summary

Plugin hook info records dropped state that the Python and JS SDKs carry, so a Java plugin could not see per-operation replay status at the attempt and change hooks, nor operation state at the invocation hooks.

**Bottom of a 2-PR stack.** The conformance handlers that assert these field shapes — and the live suite results — are in the stacked follow-up. This PR is the SDK change only and is reviewable on its own.

## What changed

**Execution payload surface (JS parity)**

- `InvocationInfo` gains `executionInput`: the deserialized input the execution was started with — the same value the durable handler receives.
- `InvocationEndInfo` gains `executionInput` (carried through from the start hook) and `executionResult`, the value the handler returned. `executionResult` is populated only when the invocation completed the execution successfully; null for PENDING/FAILED/RETRYING.
- Both are typed `Object`, matching JS's `unknown`. Note Python instead exposes `execution_result` as the **serialized** string from the invocation output; Java follows the JS shape and reports the returned object.

**Ordering change — please review.** The execution input is now deserialized *before* `onInvocationStart` fires, because the hook must carry it. Deserialization can fail, so the failure path explicitly fires the start hook with a null `executionInput` before propagating: plugins must observe an invocation-start for every invocation, including one with a malformed input payload. A regression test covers this using a `SerDes` that rejects every deserialize.

**Operation result (JS parity)**

- `result` added to the three operation-snapshot records: `OperationInfo`, `OperationEndInfo` and `OperationChangeItemInfo`.
- It is the operation's checkpointed **serialized** result, exactly as the backend recorded it, typed `String` to match JS's `result?: string`. The serialized form is deliberate — the plugin boundary has no access to the caller's target type, so it cannot deserialize. Null when the operation produced none (still running, failed, or a type carrying none such as `WAIT`).
- The payload lives on a different `*Details` member per operation type, so `BaseDurableOperation.getResultPayload` dispatches on operation type exactly as the existing `getErrorObject` does (STEP, CHAINED_INVOKE, CALLBACK, CONTEXT).

**Attempt and change-item fields**

- `UserFunctionStartInfo` / `UserFunctionEndInfo` gain `isReplay`, the operation-level indicator for whether THIS operation was observed via checkpointed state. Distinct from the existing `isReplayingChildren`, which describes the child operations of a context body and does not substitute for it.
- `OperationChangeItemInfo` was a reduced record. It now carries the full operation surface (`status`, `attempt`, `isReplay`, `error`), ordered to match `OperationEndInfo`, so an operation observed through a change delta exposes the same fields as through the per-operation hooks.

**Invocation-info enrichment**

- `InvocationInfo` gains `operations` and `updatedOperations`.
- `InvocationEndInfo` gains `operations` and `executionStartTime`. The latter was present on the start info but dropped from the end record, forcing plugins to correlate back to the start hook.
- `updatedOperations` derives from the invocation input's `UpdatedOperationIds` intersected with the tracked operations, so it is empty on the first invocation and names the externally-completed operations on a replay.
- The end-info snapshot is taken at end time, so unlike the start info it also includes operations created during the invocation.

## Design notes for reviewers

**Where the replay indicator comes from.** The attempt hook needs the operation-level replay flag at its firing site in `BaseDurableOperation.runUserFunction`, but `getOperation()` delegates to `getOperationAndUpdateReplayState`, which flips `REPLAY` to `EXECUTION` mode as a side effect. Reading it from a plugin-hook site would mutate execution state. `ExecutionManager` instead snapshots the operation ids delivered in the initial state once, and `wasObservedAtInvocationStart` is a pure containment check. That yields one consistent definition of `isReplay` across the attempt, change and invocation hooks: the operation predates this invocation.

**Map value type.** Both invocation maps are keyed by operation id and valued with `OperationChangeItemInfo`, now the richest operation snapshot record the SDK has, so a single conversion path (`PluginInfoConverter.toOperationItemMap`) feeds the change hook and both invocation hooks. The record's name is a wart in this role; renaming it to something like `OperationItemInfo` is a reasonable follow-up, kept out of this change to limit blast radius.

**No defaulting constructors.** These are positional records, so the added components surface every constructor call site at compile time. I deliberately did not add convenience overloads: a defaulting constructor is exactly how a future internal call site would silently ship empty maps to plugins, which is the failure mode this fixes. Call sites in the OpenTelemetry plugin tests were updated mechanically. All touched APIs are `@Deprecated` preview, so the record-shape changes are sanctioned.

## Out of scope

- **Requirement 10-14 / #581 — coordination needed.** 10-14 asserts an operation-end result and has a standing red-test branch (#581). This PR supplies the SDK surface that requirement needs, so #581 is expected to go green once this lands. 10-14's handler still lives on that separate branch, so it continues to report UNCOVERED in this stack's runs. Worth sequencing with whoever owns #581.
- Requirement 10-18 / #582 (Java never fires `onOperationStart` with `isReplay=true` for a pending WAIT) is a separate behavioral fix with OpenTelemetry span-model coupling. Not bundled here.

## Testing

Full module build, unit tests and `spotless:check`: BUILD SUCCESS. SDK 1129 tests, integration 400, OpenTelemetry 165, examples 120 — all green. The first commit was additionally verified applied directly to `main` with no handler code present.

Adds six `PluginIntegrationTest` cases: execution input and result on success, no execution result when suspended, none when the execution fails, the start-hook-fires-on-malformed-input invariant, and operation-end reporting the serialized result on success versus no result plus an error on failure.

Known flaky test, not caused by this PR: `InvocationOtelPluginTest.invocationEnd_closesNestedSpansChildFirst` (a test added on `main`) failed 3 times across ~12 full-reactor runs on this branch, and 0 times across 8 runs on pristine `main`; it passes 6/6 in isolation on both. This PR changes **no** OpenTelemetry production code (`git diff main HEAD -- otel-plugin/src/main` is empty) and the span close order in `endOpenSpansChildFirst` is deterministic, so the correlation looks like timing sensitivity in the test rather than a behaviour change. Flagging rather than hiding it.

Conformance evidence lives in the stacked follow-up, where the handlers exist to assert these shapes: 21/21 covered plugin requirements pass, with 10-19, 10-21 and 10-22 flipping to green.

Refs: #604

### DIFF
--- a/sdk/src/main/java/software/amazon/lambda/durable/execution/DurableExecutor.java
+++ b/sdk/src/main/java/software/amazon/lambda/durable/execution/DurableExecutor.java
@@ -28,6 +28,7 @@ import software.amazon.lambda.durable.model.DurableExecutionOutput;
 import software.amazon.lambda.durable.plugin.InvocationEndInfo;
 import software.amazon.lambda.durable.plugin.InvocationInfo;
 import software.amazon.lambda.durable.plugin.InvocationStatus;
+import software.amazon.lambda.durable.plugin.PluginInfoConverter;
 import software.amazon.lambda.durable.plugin.PluginRunner;
 import software.amazon.lambda.durable.serde.SerDes;
 import software.amazon.lambda.durable.util.ExceptionHelper;
@@ -67,11 +68,19 @@ public class DurableExecutor {
                         // onInvocationStart runs on the user thread so plugins can
                         // inject ThreadLocal objects, update MDC, etc.
                         // executionStartTime comes from the initial EXECUTION operation in the first backend event.
+                        // The operation maps are snapshots of the state delivered for this invocation; on a replay
+                        // invocation updatedOperations names the operations the backend completed while suspended.
                         pluginRunner.onInvocationStart(new InvocationInfo(
                                 requestId,
                                 executionArn,
                                 isFirstInvocation,
-                                executionManager.getExecutionOperation().startTimestamp()));
+                                executionManager.getExecutionOperation().startTimestamp(),
+                                PluginInfoConverter.toOperationItemMap(
+                                        executionManager.getOperationsSnapshot(),
+                                        executionManager.getInitialOperationIds()),
+                                PluginInfoConverter.toOperationItemMap(
+                                        executionManager.getUpdatedOperationsSnapshot(),
+                                        executionManager.getInitialOperationIds())));
 
                         var userInput = extractUserInput(
                                 executionManager.getExecutionOperation(), config.getSerDes(), inputType);
@@ -99,6 +108,7 @@ public class DurableExecutor {
                                 if (cause instanceof SuspendExecutionException) {
                                     fireOnInvocationEnd(
                                             pluginRunner,
+                                            executionManager,
                                             requestId,
                                             executionArn,
                                             isFirstInvocation,
@@ -115,6 +125,7 @@ public class DurableExecutor {
                                         && unrecoverableDurableExecutionException.isRetryable()) {
                                     fireOnInvocationEnd(
                                             pluginRunner,
+                                            executionManager,
                                             requestId,
                                             executionArn,
                                             isFirstInvocation,
@@ -127,6 +138,7 @@ public class DurableExecutor {
                                 logger.debug("Execution failed: {}", cause.getMessage());
                                 fireOnInvocationEnd(
                                         pluginRunner,
+                                        executionManager,
                                         requestId,
                                         executionArn,
                                         isFirstInvocation,
@@ -141,6 +153,7 @@ public class DurableExecutor {
                                     DurableExecutionOutput.success(handleLargePayload(executionManager, outputPayload));
                             fireOnInvocationEnd(
                                     pluginRunner,
+                                    executionManager,
                                     requestId,
                                     executionArn,
                                     isFirstInvocation,
@@ -159,12 +172,24 @@ public class DurableExecutor {
 
     private static void fireOnInvocationEnd(
             PluginRunner pluginRunner,
+            ExecutionManager executionManager,
             String requestId,
             String executionArn,
             boolean isFirstInvocation,
             InvocationStatus status,
             Throwable error) {
-        pluginRunner.onInvocationEnd(new InvocationEndInfo(requestId, executionArn, isFirstInvocation, status, error));
+        // The end info repeats the start info's identity surface (execution start time, operation snapshot) so an
+        // invocation-end hook never has to correlate back to the start hook. The snapshot is taken at end time, so
+        // unlike the start info it also contains operations created during this invocation.
+        pluginRunner.onInvocationEnd(new InvocationEndInfo(
+                requestId,
+                executionArn,
+                isFirstInvocation,
+                executionManager.getExecutionOperation().startTimestamp(),
+                PluginInfoConverter.toOperationItemMap(
+                        executionManager.getOperationsSnapshot(), executionManager.getInitialOperationIds()),
+                status,
+                error));
     }
 
     private static String handleLargePayload(ExecutionManager executionManager, String outputPayload) {
--- a/sdk/src/main/java/software/amazon/lambda/durable/execution/DurableExecutor.java
+++ b/sdk/src/main/java/software/amazon/lambda/durable/execution/DurableExecutor.java
@@ -29,6 +29,7 @@ import software.amazon.lambda.durable.model.DurableExecutionOutput;
 import software.amazon.lambda.durable.plugin.InvocationEndInfo;
 import software.amazon.lambda.durable.plugin.InvocationInfo;
 import software.amazon.lambda.durable.plugin.InvocationStatus;
+import software.amazon.lambda.durable.plugin.PluginInfoConverter;
 import software.amazon.lambda.durable.plugin.PluginRunner;
 import software.amazon.lambda.durable.serde.SerDes;
 import software.amazon.lambda.durable.util.ExceptionHelper;
--- a/sdk/src/main/java/software/amazon/lambda/durable/execution/DurableExecutor.java
+++ b/sdk/src/main/java/software/amazon/lambda/durable/execution/DurableExecutor.java
@@ -87,18 +87,20 @@ public class DurableExecutor {
                         // onInvocationStart runs on the user thread so plugins can
                         // inject ThreadLocal objects, update MDC, etc.
                         // executionStartTime comes from the initial EXECUTION operation in the first backend event.
-                        pluginRunner.onInvocationStart(new InvocationInfo(
-                                requestId,
-                                executionArn,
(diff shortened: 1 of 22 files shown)


sdk/src/main/java/software/amazon/lambda/durable/plugin/OperationChangeItemInfo.java
```diff
@@ -27,5 +30,17 @@ public record OperationChangeItemInfo(
         String parentId,
         Instant startTimestamp,
         Instant endTimestamp,
+        OperationStatus status,
+        Integer attempt,
+        boolean isReplay,
```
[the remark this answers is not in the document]
wangyb-A: the plugin interface is not officialy released, it's safe to change the args

sdk/src/main/java/software/amazon/lambda/durable/plugin/PluginInfoConverter.java
```diff
@@ -104,19 +106,25 @@ private static String extractResult(Operation operation) {
...
+     * @param isReplay true if this operation was already present in the checkpointed state when it started
+     * @param isReplayingChildren true if the child operations of this context body are replaying from checkpoints
      * @param attempt the 1-based attempt number (null for context operations)
      * @return a UserFunctionStartInfo record
      */
     public static UserFunctionStartInfo toUserFunctionStartInfo(
-            OperationIdentifier identifier, String parentId, boolean isReplayingChildren, Integer attempt) {
+            OperationIdentifier identifier,
+            String parentId,
+            boolean isReplay,
```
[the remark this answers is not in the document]
wangyb-A: the plugin interface is not officialy released, it's safe to change the args

sdk/src/main/java/software/amazon/lambda/durable/execution/DurableExecutor.java
```diff
@@ -86,12 +87,20 @@ public static <I, O> DurableExecutionOutput execute(
                         // onInvocationStart runs on the user thread so plugins can
                         // inject ThreadLocal objects, update MDC, etc.
                         // executionStartTime comes from the initial EXECUTION operation in the first backend event.
-                        pluginRunner.onInvocationStart(new InvocationInfo(
-                                requestId,
-                                executionArn,
-                                isFirstInvocation,
-                                executionManager.getExecutionOperation().startTimestamp(),
-                                userInput));
+                        if (!pluginRunner.isEmpty()) {
```
zhongkechen, line 90: nit: empty check can be on the pluginRunner side.

sdk/src/main/java/software/amazon/lambda/durable/plugin/PluginInfoConverter.java
```diff
@@ -104,19 +106,25 @@ private static String extractResult(Operation operation) {
      *
      * @param identifier the operation identifier containing id, name, type, and subType
      * @param parentId the parent operation ID (may be null)
-     * @param isReplay true if the user function is called during replay (context operations)
+     * @param isReplay true if this operation was already present in the checkpointed state when it started
```
zhongkechen: Confusing. Can we find a better name and explain the behavior better?

sdk/src/main/java/software/amazon/lambda/durable/plugin/PluginInfoConverter.java
```diff
@@ -104,19 +106,25 @@ private static String extractResult(Operation operation) {
...
+     * @param isReplayingChildren true if the child operations of this context body are replaying from checkpoints
```
zhongkechen: Confusing. What's the difference from above? Can we find a better name and explain the behavior better?

sdk/src/main/java/software/amazon/lambda/durable/plugin/PluginInfoConverter.java
```diff
@@ -153,25 +162,41 @@ public static UserFunctionEndInfo toUserFunctionEndInfo(
...
+     *     each item's {@code isReplay} indicator
      * @return an OperationChangeInfo record
      */
     public static OperationChangeInfo toOperationChangeInfo(
             String requestId,
             String durableExecutionArn,
             Collection<Operation> updatedOperations,
-            Collection<Operation> allOperations) {
+            Collection<Operation> allOperations,
+            Set<String> replayedOperationIds) {
```
zhongkechen: Confusing name. `initialOperationIds` sounds better?

sdk/src/main/java/software/amazon/lambda/durable/execution/DurableExecutor.java
wangyb-A: This check avoid to copy the operations from the execution manager, if we do that on plugin runner side, we have to pass the execution manager to it

sdk/src/main/java/software/amazon/lambda/durable/plugin/PluginInfoConverter.java
wangyb-A: only keep isReplay to match js

sdk/src/main/java/software/amazon/lambda/durable/plugin/PluginInfoConverter.java
wangyb-A: only keep isReplay to match js

sdk/src/main/java/software/amazon/lambda/durable/plugin/UserFunctionStartInfo.java
```diff
@@ -26,5 +26,5 @@ public record UserFunctionStartInfo(
         String subType,
         String parentId,
         Instant startTimestamp,
-        boolean isReplayingChildren,
+        boolean isReplay,
```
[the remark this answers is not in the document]
wangyb-A: we are still in preview, no one is actively using isReplayingChildren, and we want to match the interface with other sdks

sdk/src/main/java/software/amazon/lambda/durable/plugin/PluginInfoConverter.java
zhongkechen: The interface hasn't released for js yet. We can rename for both Java and JS if we have a better name

sdk/src/main/java/software/amazon/lambda/durable/plugin/PluginInfoConverter.java
zhongkechen: Same as the above. `isReplayingChildren` could be renamed.

sdk/src/main/java/software/amazon/lambda/durable/plugin/PluginInfoConverter.java
wangyb-A: I think we can keep this name `isReplay`, since this apply to all operations, not limited to child context operation, do you have any suggession on this?

zhongkechen (approved):

PR #673, opened 2026-09-01
Round-trip WaitForCondition's initialState through SerDes on start()

### LINKED ISSUE #671: WaitForCondition: round-trip initialState through SerDes before first check
In the Python SDK, wait_for_condition round-trips initial_state through the configured serdes before the first check runs, so the check sees the same serialized-then-deserialized shape on every poll. The Java SDK currently passes config.initialState() raw into the first check and only normalizes the check result through SerDes afterward, so the first poll's input state differs in shape from subsequent polls (which are deserialized from the checkpoint).

### Current behavior
In WaitForConditionOperation.java:
```
@Override
protected void start() {
    executeCheckLogic(config.initialState(), FIRST_ATTEMPT);
}
```
start() passes config.initialState() directly. executeCheckLogic normalizes the check result via serializeAndDeserializeResult(result.value()), and resumeCheckLoop deserializes currentState from the checkpoint - but the initial state itself is never round-tripped before the first check.

### Proposed behavior
Round-trip config.initialState() through the configured SerDes before invoking the first check, so the check observes a consistent shape on every attempt and non-serializable initial state fails fast.

Updated docs repo with the change as well.

KaioAntonio:
Fixes #671

## Root cause

`WaitForConditionOperation.start()` passed `config.initialState()` straight into the first check:

```java
@Override
protected void start() {
    executeCheckLogic(config.initialState(), FIRST_ATTEMPT);
}
```

Every subsequent attempt goes through `resumeCheckLoop`, which derives `currentState` by deserializing either the checkpoint payload or, when there's no checkpoint yet, `config.initialState()` itself:

```java
T currentState;
if (checkpointData != null) {
    currentState = deserializeResult(checkpointData);
} else {
    currentState = config.initialState();
}
```

So every check after the first sees a value that has gone through the configured `SerDes`, but the very first check sees the raw, un-normalized value. A `SerDes` that changes shape on round-trip (narrows a type, drops unknown fields, changes floating-point precision, etc.) makes the first invocation observe a different shape than every later one for the exact same logical state — matching the inconsistency this issue describes relative to the Python SDK's `wait_for_condition`.

## Fix

`SerializableDurableOperation` already has `serializeAndDeserializeResult()`, documented for exactly this purpose ("keeps first execution consistent with replay when a SerDes normalizes or otherwise changes the value") and already used in `executeCheckLogic` to normalize the check *result*. `start()` now uses the same helper on the initial state before the first check:

```java
@Override
protected void start() {
    var initialState = serializeAndDeserializeResult(config.initialState()).deserialized();
    executeCheckLogic(initialState, FIRST_ATTEMPT);
}
```

## Testing

- Added `startNormalizesInitialStateThroughSerDes` to `WaitForConditionOperationTest`, using a `SerDes` mock where `deserialize` returns a value distinct from the raw `initialState`, to prove the check function receives the round-tripped value rather than the raw one. Confirmed this test fails on the unfixed code (`expected: <99> but was: <42>`) and passes with the fix.
- `mvn -pl sdk -am test`: full `sdk` module suite passes, no failures.
- `mvn -pl sdk spotless:check`: clean.

### DIFF
--- a/sdk/src/test/java/software/amazon/lambda/durable/operation/WaitForConditionOperationTest.java
+++ b/sdk/src/test/java/software/amazon/lambda/durable/operation/WaitForConditionOperationTest.java
@@ -7,7 +7,9 @@ import static org.mockito.Mockito.*;
 
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
@@ -31,6 +33,7 @@ import software.amazon.lambda.durable.model.OperationIdentifier;
 import software.amazon.lambda.durable.model.OperationSubType;
 import software.amazon.lambda.durable.model.WaitForConditionResult;
 import software.amazon.lambda.durable.serde.JacksonSerDes;
+import software.amazon.lambda.durable.serde.SerDes;
 
 class WaitForConditionOperationTest {
 
@@ -371,13 +374,14 @@ class WaitForConditionOperationTest {
 
     @Test
     void startNormalizesInitialStateThroughSerDes() throws Exception {
-        var mockSerDes = mock(software.amazon.lambda.durable.serde.SerDes.class);
+        var mockSerDes = mock(SerDes.class);
         when(mockSerDes.serialize(42)).thenReturn("42-normalized");
         when(mockSerDes.deserialize(eq("42-normalized"), any())).thenReturn(99);
 
         when(executionManager.getOperationAndUpdateReplayState(OPERATION_ID)).thenReturn(null);
 
         var receivedState = new java.util.concurrent.atomic.AtomicInteger(-1);
+        var checkRan = new CountDownLatch(1);
         var config = WaitForConditionConfig.<Integer>builder()
                 .serDes(mockSerDes)
                 .initialState(42)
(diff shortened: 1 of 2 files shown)


sdk/src/test/java/software/amazon/lambda/durable/operation/WaitForConditionOperationTest.java
```diff
@@ -367,6 +367,38 @@ void replayStartedWithNullCheckpointDataUsesInitialState() throws Exception {
...
+        var operation = createOperation(
+                (state, ctx) -> {
+                    receivedState.set(state);
+                    return WaitForConditionResult.stopPolling(state);
+                },
+                config);
+
+        operation.execute();
+
+        Thread.sleep(200);
```
zhongkechen: nit: replacing the new Thread.sleep(200) test wait with a latch would reduce flakiness

sdk/src/test/java/software/amazon/lambda/durable/operation/WaitForConditionOperationTest.java
```diff
@@ -367,6 +367,38 @@ void replayStartedWithNullCheckpointDataUsesInitialState() throws Exception {
         assertEquals(42, receivedState.get(), "Should use initialState when checkpoint data is null");
     }
 
+    // ===== start() SerDes round-trip =====
+
+    @Test
+    void startNormalizesInitialStateThroughSerDes() throws Exception {
```
zhongkechen, line 376: Add coverage for withDeserializeAfterSerialization(false) to confirm serialization still occurs but the check receives the raw initial state.

KaioAntonio:
Thanks @zhongkechen, both addressed in ec759c6:

- Replaced the `Thread.sleep(200)` with a `CountDownLatch` the check function counts down, so the test doesn't depend on a fixed delay.
- Added `startSerializesInitialStateButSkipsDeserializeWhenDisabled`: with `withDeserializeAfterSerialization(false)`, `serialize()` still runs against the initial state (so a non-serializable value still fails fast), but `deserialize()` is never called and the check receives the raw `initialState` rather than a round-tripped value.

`mvn -pl sdk -am test` and `spotless:check` both pass. (The `e2e-tests`/`conformance` CI failures are unrelated — those need AWS credentials that GitHub doesn't grant to fork PRs, so they fail the same way for any external contributor; `build (17/21/25)` and `spotless-check`, which run the actual unit suite, are green.)

zhongkechen (approved):

Issue #679, opened 2026-09-02
[Feature]: Java Workflow Insight plugin follow-up work


wangyb-A:
### What would you like?

Add a Workflow Insight plugin to the AWS Lambda Durable Execution SDK for Java.

The plugin should emit a structured summary of each durable execution, including
execution status, timing, input/output, errors, and operation details. It should
follow the JavaScript Workflow Insight contract where practical while accounting
for Java-specific threading, serialization, and plugin lifecycle behavior.

### Why?

Customers need an easier way to inspect and analyze durable workflows without
reconstructing execution state from raw Lambda logs and execution history.

Workflow Insight records enable:

- Execution monitoring and troubleshooting
- Workflow latency and failure analysis
- Operation-level visibility
- Querying records through CloudWatch Logs, S3, and analytics tools
- Consistent observability across JavaScript, Python, and Java SDKs

### Initial scope

- Add a publishable `insight-plugin` Maven module.
- Support schema version `1.0` of the Workflow Insight record.
- Support `ON_COMPLETE`, `ON_FAILURE`, and `ON_CHANGE` emission modes.
- Support deterministic per-execution sampling.
- Support top-level and full-tree operation detail.
- Support input/output transforms, operation result transforms, filtering, and
  error controls.
- Support record-size limits and deterministic truncation.
- Isolate plugin and exporter failures from durable execution.
- Publish the plugin to Maven Central and attach it to GitHub releases.

### Acceptance criteria

- [ ] Java plugin hooks expose the execution and operation data required by the
      record contract.
- [ ] Workflow Insight module builds as part of the Maven reactor.
- [ ] Records include execution metadata, operation summaries, errors, attempts,
      and opted-in results.
- [ ] S3 emits the canonical `operations` array.
- [ ] Lambda and CloudWatch Logs emit `operationsByName`.
- [ ] Sampling, content controls, filtering, truncation, and exporter isolation
      have unit coverage.
- [ ] S3 Workflow Insight conformance passes all applicable requirements.
- [ ] CloudWatch Workflow Insight conformance passes all applicable requirements.
- [ ] Release automation publishes the new Maven artifact.
- [ ] Public preview APIs use `@Experimental`.
- [ ] `includeErrors(false)` suppresses execution and operation errors.
- [ ] Plugin `Throwable`s do not disrupt durable execution.
- [ ] Add an ADR documenting the record contract, lifecycle, sampling,
[...]

PR #701, opened 2026-09-10
feat(insight): add SQSExporter

wangyb-A:

### Issue Link, if available

Related: #679

### Description

Adds `SQSExporter` to the Workflow Insight plugin, so insight records can be shipped to an Amazon SQS queue. It is modeled on the existing `S3Exporter` and `CloudWatchLogsExporter` and mirrors the official JavaScript `SQSExporter` contract where practical. No change was made to the SDK core, `WorkflowInsight`, or `WorkflowInsightRecord`.

**Standard vs FIFO queues.** The queue type is detected from the URL: a `queueUrl` ending in `.fifo` is treated as FIFO. For a standard queue the `SendMessage` request carries no `MessageGroupId` and no `MessageDeduplicationId`. For a FIFO queue both are always set.

**FIFO id hashing, bounds, and dedup.** SQS caps both ids at 128 characters. The `MessageGroupId` is the explicit `messageGroupId` (else the `executionArn`) used verbatim when it is 128 characters or shorter, otherwise a deterministic 64-character SHA-256 hex digest of that raw value, so grouping stays stable per execution. The `MessageDeduplicationId` is always a 64-character SHA-256 digest of `executionArn \0 emittedAt \0 renderedBody`. Two different snapshots at the same `emittedAt` differ by body, so they get different dedup ids and both survive the SQS 5-minute dedup window; an exact retry has the same body and the same id, so SQS de-duplicates it. Hashing uses the JDK `java.security.MessageDigest`, no new dependency.

**Operation formats.** `operationsFormat` mirrors the JS enum: `ARRAY` (default), `BY_NAME`, and `BOTH`. `render()` produces the matching wire map. `BOTH` is built inside the exporter from the record's public API, so `WorkflowInsightRecord` needed no new method.

**Optional dependency.** The `sqs` Maven dependency is declared `<optional>true</optional>` with its version inherited from the imported AWS SDK BOM, matching the `s3` and `cloudwatchlogs` treatment, so default Lambda-log consumers do not inherit it.

**IAM docs.** The README documents the required `sqs:SendMessage` permission and FIFO queue setup alongside the usage and options.

**Size headroom.** `maxRecordSizeBytes` defaults to `256000` and bounds the rendered JSON body only. SQS counts message attributes toward the same 256 KB quota, so the default is left below the quota to leave attribute headroom. `build()` rejects a non-positive `maxRecordSizeBytes` (which previously disabled truncation silently).

### Demo/Screenshots

N/A -- backend-only change, no user-visible UI.

### Checklist

#### Unit Tests

Yes. `SQSExporterTest` has 13 tests covering: standard-queue requests omit FIFO fields; FIFO default group id is the `executionArn` and dedup id is derived from `executionArn`/`emittedAt`/body; an explicit `messageGroupId` is honored; a long group id is bounded to a stable SHA-256 hash; rapid same-timestamp snapshots get distinct dedup ids while exact retries de-duplicate; `build()` rejects a non-positive `maxRecordSizeBytes`; and `BY_NAME` and `BOTH` render shapes still derive the bounded dedup id from `emittedAt`. Insight-plugin reactor: 82 tests, 0 failures. Full reactor `mvn install`: BUILD SUCCESS, 1954 tests, 0 failures, 31 skipped (cloud tests disabled by default).

#### Integration Tests

None added. The exporter is an isolated component fully covered by mocked-client unit tests, consistent with `S3ExporterTest` and `CloudWatchLogsExporterTest`, which are also unit-only.

#### Examples

No new example added.

### DIFF
--- a/insight-plugin/README.md
+++ b/insight-plugin/README.md
@@ -28,7 +28,8 @@ DurableConfig config = DurableConfig.builder()
 
 Exporters: `LambdaLogExporter` (default; writes the `operationsByName` map to stdout →
 CloudWatch), `S3Exporter` (canonical `operations` array, one object per execution),
-`CloudWatchLogsExporter` (PutLogEvents to a specific log group, `operationsByName` map). Implement
+`CloudWatchLogsExporter` (PutLogEvents to a specific log group, `operationsByName` map),
+`SQSExporter` (one `SendMessage` per emission; standard or FIFO queue). Implement
 `InsightExporter` for custom sinks.
 
 `LambdaLogExporter` needs no extra dependency. The AWS SDK service modules used by the remote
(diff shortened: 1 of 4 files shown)


wangyb-A:
**Intent:** Add an `SQSExporter` to the Workflow Insight plugin so insight records can be delivered to an Amazon SQS queue (standard and FIFO), mirroring the official JavaScript exporter contract where practical, with SQS-safe bounded FIFO group/dedup ids.
**Not a goal:** Changing the SDK core, `WorkflowInsight`, or `WorkflowInsightRecord`; adding integration/end-to-end tests; or making the `sqs` dependency non-optional.

PR #702, opened 2026-09-10
feat(insight): add HttpExporter

wangyb-A:

### Issue Link, if available

Related: #679

### Description

This adds `HttpExporter` to the Workflow Insight plugin. It sends one insight record to an HTTP endpoint as one request. It mirrors the JS `HttpExporter` so the two SDKs behave the same.

What it does:

- Sends the record with `POST` or `PUT`. `POST` is the default.
- Sets `Content-Type: application/json`. A caller may pass their own headers, including their own `Content-Type`. The caller's value wins. Header names are matched without caring about upper or lower case, so a lowercase `content-type` replaces the default instead of adding a second one. The endpoint receives exactly one `Content-Type` header.
- Waits at most `timeoutMs` for the endpoint. The default is 10000 ms. A non-2xx reply throws.
- Shapes the body with `operationsFormat`: `ARRAY` (the default) writes the `operations` array, `BY_NAME` writes the `operationsByName` map, `BOTH` writes both. `render()` builds the shape and `export()` sends it, sharing one path.

How it talks HTTP:

- It uses the JDK's own `java.net.http.HttpClient`. This adds no new dependency. The client sits behind a small package-visible `HttpSender` seam so tests can inject a fake, the same way `S3Exporter` and `CloudWatchLogsExporter` allow a `.client(...)`.
- It never copies the server's response body into a log. A failure throws with the numeric status and a derived reason label only. This is on purpose: an arbitrary response body could carry sensitive server content.

No change to the SDK core or the Workflow Insight plugin wiring.

### Demo/Screenshots

N/A. This is a backend Java change with no user-visible UI, so there is nothing to screenshot.

### Checklist

#### Unit Tests

Yes. `HttpExporterTest` (17 tests) covers, with a mock sender: default `POST` + `Content-Type` + array body + 10s timeout; `PUT` with a merged `Authorization` header; non-2xx throws; `BY_NAME` map; `BOTH` array + map; a configured timeout is applied; the caller's `Content-Type` overrides the default case-insensitively with no duplicate; and builder rejects a missing url, a non-http scheme, a non-absolute url, and a non-positive timeout.

#### Integration Tests

Transport-level tests run the real default JDK sender end to end against an in-process `com.sun.net.httpserver.HttpServer`: a `POST` returns 200 and the handler asserts method, `Content-Type`, and body; a 500 throws; a `PUT` with a custom `Authorization` header is received; the wire carries exactly one `Content-Type` when the caller overrides its casing; a connection to a dead loopback port throws `IllegalStateException`; and an interrupted export throws while restoring the thread's interrupt flag.

#### Examples

N/A. No new example is needed for a new exporter class; the plugin README's HttpExporter section documents customer usage.

### DIFF
--- a/insight-plugin/README.md
+++ b/insight-plugin/README.md
@@ -55,9 +55,9 @@ configure, using the AWS SDK for Java 2.x version managed by your application:
 
 ### HttpExporter
 
-`HttpExporter` POSTs (or PUTs) each record to any HTTP(S) endpoint or webhook as a JSON body with
-`Content-Type: application/json`. It uses the JDK's built-in HTTP client, so it adds **no dependency**.
-It mirrors the JS `HttpExporter`.
+`HttpExporter` POSTs (or PUTs) each record to any HTTP(S) endpoint or webhook as a JSON body with a
+default `Content-Type: application/json` header that a caller may override. It uses the JDK's built-in
+HTTP client, so it adds **no dependency**. It mirrors the JS `HttpExporter`.
 
 ```java
 .addExporter(HttpExporter.builder()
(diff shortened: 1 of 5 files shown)


wangyb-A:
**Intent:** Add an `HttpExporter` to the Workflow Insight plugin that POSTs/PUTs one insight record as JSON to a generic HTTP endpoint, mirroring the JS SDK's HttpExporter, using only the JDK HttpClient (no new dependency).
**Not a goal:** No retry/backoff, no response-body logging, no SDK-core or Workflow Insight wiring changes, and no promotion of a `toBothWireMap()` to core.

